PR #20, opened 2024-06-20
[ CLEANUP] Deprecate some config args that are not needed

LRL-ModelCloud:


Qubitium:
Note: We need to investigate compatibility with old autogptq quants before I can merge this. Need to make sure all 3rd quants by are not broken by this commit. 

TODO
* [ ] Old quant compat on HF. Find some older quants and check the config.json and see if they load correctly using our main.

PR #24, opened 2024-06-20
[MODEL] Add qwen moe support

LRL-ModelCloud:


Qubitium:
@bozheng-hit  We have refractored your code to be more generic so that the same `dynamic_expert_index` variable expansion code for model defs can be used by future moe model as well. Currently we are quantiing and doing validation to make sure ppl before and after quant is in normal range. I do not see major problems we should have this merged in the next couple of hours.

Qubitium:
TEST PASSED:

Native PPL: 6.4653
Quant PPL: 6.8459

PR #38, opened 2024-06-21
[MODEL] Add dbrx converted support

LRL-ModelCloud:


gptqmodel/models/dbrx.py
```diff
@@ -0,0 +1,5 @@
+from .base import BaseGPTQModel
+
+
+class DbrxGPTQ(BaseGPTQModel):
+    info = {"notes": "Dbrx is only supported via special defused/converted models in hf.co/ModelCloud repo with trust_remote_enabled"}
```
Qubitium: should be `info.notes = ` here. No need to override the entire dict from base. Also need to add `supported` boolean default True to BaseGPTQModel and here set `supported = False`.  and check on `from_pretrained` if .supported is false, raise UnsupportedError...

Qubitium:
@LRL-ModelCloud Everything looks good to me. Once the converted models are uploaded to hf.co/ModelCloud, I will merge this.

Qubitium (approved):

PR #39, opened 2024-06-21
[CORE] Add Bitblas Format/Kernel Support

ZX-ModelCloud:


Qubitium:
@ZX-ModelCloud  rename BaseQuantizeConfig to QuantizeConfig. BaseQuantizeConfig is only used for backward compat with old AutoGPTQ users.

Qubitium:
@ZX-ModelCloud  Make sure all new methods are properly type hinted.

Qubitium:
@LeiWang1999  

1. I have pushed PR to fix Bitblas as there is bad/incomplete support for gpu detection via nvidia-smi. 
2. Using tinyllama quantized as test, the bitblas conversion and for the first layer it takes about 300s to convert and 2nd layer forward becomes very fast. is TVM caching/bitblas JIT compiling the code the layer 0 and reusing for subsequent layers?

For now, we will monkeypatch bitblas so we can get it working with this PR in our multi-gpu env. 

Thanks!

LeiWang1999:
Hi @Qubitium, thank you for your attention. BitBlas will take several minutes for compilation for dynamic shapes when encountering new operator configurations. The tuning results will be saved by default in the database located at `~/.cache/bitblas`. Therefore, when layer 0 is initialized next time, the operator should load directly from the database, bypassing the 5-minute wait.

And feel free to make a pull request to BitBLAS for the patch of multi-gpus' detection!

Qubitium:
> And feel free to make a pull request to BitBLAS for the patch of multi-gpus' detection!

PR created https://github.com/microsoft/BitBLAS/pull/58

We will do some benchmarks on quantized llama7-8b model. If it beats out marlin, exllama v1/v2, we will use as default =)

LeiWang1999:
@Qubitium Thank you for the detailed feedback and review. Regarding your query on the Bitblas disk format, I can confirm that the current BITBLAS disk format is stable, and there are no immediate plans to change or modify the format.

Moreover, bitblas maintains the interface ` global_operator_cache.load_from_database(
                BITBLAS_DATABASE_PATH, BITBLAS_TARGET
            )`, in the event of any updates to the bitblas disk format in future release, we will update the warning message to reflect these changes as well.

Qubitium:
>  I can confirm that the current BITBLAS disk format is stable, and there are no immediate plans to change or modify the format.

@ZX-ModelCloud  We are safe to move forward with refractor and add `FORMAT.BITBLAS`, This would not only simplify logic but actually allows much more flexibility for how bitblas is integrated moving forward.

LeiWang1999:
Thanks for all of your hard efforts! @Qubitium @ZX-ModelCloud

PR #41, opened 2024-06-21
[CI] allow run the specified test files

CSY-ModelCloud:


Qubitium:
List of test files to run (default all) => Input Test to Run (default all)
remove .py from all test names.

PR #44, opened 2024-06-21
[CORE] Disable exllamav1 for inference if v1 if slower than v2

CSY-ModelCloud:


Qubitium:
Also need to benchmark on 4090

Llama7B

1. batch 1 input 128
2. batch 4 input 128
3. batch 1 input 512
4. batch 4 input 512

Exllama v1 vs v2 vs Marlin. 

symTrue desc-actFalse

Issue #59, opened 2024-06-25
[FEATURE] Consolidate 6+ related `use/disable: bool` args in `from_quantized` into single `backend: Backend`


Qubitium:
The following args will be merged into single `backed: Backend = Backend.AUTO`

* use_triton: bool,
* disable_exllama: bool = False,
* disable_exllamav2: bool = False,
* use_marlin: bool = False,
* use_bitblas: bool = True,

Reason: It is not only super confusing for users to use correctly (matrix condition of passive binary toggles), even project developers ran into multiple bugs due to these passive switches. We can't keep adding more binary toggles every time we add a backend/kernel/runtime. Becoming unmaintainable and unusable by both end-users and project devs.

Prelim design:

```py
class Backend(ENUM):
   AUTO # choose the fastest one based on quant model compatibility
   CUDA_OLD
   CUDA 
   TRITON_V2
   EXLLAMA
   EXLLAMA_V2
   MARLIN
   BITBLAS
 ```

This is the final todo for 0.9.1 Release.

Qubitium:
Update: `use_triton` was incorrectly managed. It is actually a overlay format (sub-format) of gptq v1/v2 and not a dynamic kernel overly on top of pure gptq v1/v2 format as we had previously assumed.  We are going to fix/refractor so that GPTQ_TRITON and GPTQ_V2_TRITON formats are correctly used instead of the bad `use_triton` boolean hint.

PR #68, opened 2024-06-26
Consolidate Backend

ZX-ModelCloud:
Resolves https://github.com/ModelCloud/GPTQModel/issues/59

The following args will be merged into single backed: Backend = Backend.AUTO

use_triton: bool,
disable_exllama: bool = False,
disable_exllamav2: bool = False,
use_marlin: bool = False,
use_bitblas: bool = True,
Reason: It is not only super confusing for users to use correctly (matrix condition of passive binary toggles), even project developers ran into multiple bugs due to these passive switches. We can't keep adding more binary toggles every time we add a backend/kernel/runtime. Becoming unmaintainable and unusable by both end-users and project devs.

Prelim design:

class Backend(ENUM):
   AUTO # choose the fastest one based on quant model compatibility
   CUDA_OLD
   CUDA 
   TRITON_V2
   EXLLAMA
   EXLLAMA_V2
   MARLIN
   BITBLAS


Qubitium:
Update: use_triton was incorrectly managed. It is actually a overlay format (sub-format) of gptq v1/v2 and not a dynamic kernel overly on top of pure gptq v1/v2 format as we had previously assumed. We are going to fix/refractor so that GPTQ_TRITON and GPTQ_V2_TRITON formats are correctly used instead of the bad use_triton boolean hint.

gptqmodel/utils/importer.py
```diff
@@ -1,35 +1,73 @@
...
+        format: str,
+        pack: bool = False,
 ):
-    # TODO handle AUTO
+    # Handle the case where backend is AUTO.
+    if backend == Backend.AUTO:
+        allow_backends = format_dict[format]
+        for k, v in backend_dict.items():
+            in_allow_backends = k in allow_backends if pack else True
+            validate = v.validate(bits, group_size, desc_act, sym, raise_error=True)
```
Qubitium: I think you need to change raise_error=False so it doesn't crash here?

ZX-ModelCloud: updated.

PR #81, opened 2024-06-27
[CI] Reduce execution time of unit tests

LRL-ModelCloud:


Qubitium:
@LRL-ModelCloud  Also add a new Batch unit test `test_quant_batch.py` comparing no batch (batch_size = 1) vs batch (batch_size = N) and make sure the two PPL post quantization is within < 0.X small range.

PR #98, opened 2024-06-28
[FIX] Padding infeatures/outfeatures for exllama, exllama v2, and marlin

### LINKED ISSUE #100: [BUG] Padding of infeatures/outfeatures and packing
Exllama v1/v2 + Marlin have (or will have) auto padding of infeatures and outfeatures so many models can run despite not exactly matching kernel restrictions. 

However, if the padding is done dynamically at runtime, and we pack the padded tensors, the result weights may not be compatible with vllm/sglang as the shape of the tensors are off due to padding. 

This is a theory at the moment and needs investigation. If confirmed, we need to trim the tensors during or post packing to original infeatures/outfeatures shape/size/

Qubitium:
Resolves #100


Qubitium:
Confirmed my suspicion that padding code will impact the shape of saved tensors to disk. Pack has been fixed so that the saved size is correct (original) but now running into load issues since the expanded/padded buffers are now larger than the tensors on disk so accelerate is throwing shape not same errors. We will explore 2 methods to deal with this tomorrow.

Qubitium:
Update: We are going with Plan B.

PR #103, opened 2024-06-28
[REFRACTOR] gptqmodel_post_init 

PZS-ModelCloud:


gptqmodel/models/base.py
```diff
@@ -1030,10 +1030,6 @@ def skip(*args, **kwargs):
             quantize_config.format = FORMAT.GPTQ_V2
 
         if backend == Backend.MARLIN:
-            if is_sharded:
```
Qubitium: why is this removed?

Issue #117, opened 2024-06-29
[BUG] OPT model and fc1/fc2 modules


Qubitium:
From our testing, OPT model has suboptimal model config where fc1/fc2 modules should not be quantized as they are not a good candidate for quantization due to massive losses during calibration. 

Investigate and confirm.

PR #118, opened 2024-06-29
[FIX] opt fc1/fc2 layer modules should not be quantized

Qubitium:
Resolves https://github.com/ModelCloud/GPTQModel/issues/117

For OPT model, the fc1/fc2 layer modules are horrible/in-compatible with current quant calibration resulting in massive losses for every single layer.  Solution is to disable/skip these 2 modules.

### DIFF
--- a/gptqmodel/models/opt.py
+++ b/gptqmodel/models/opt.py
@@ -15,6 +15,6 @@ class OPTGPTQ(BaseGPTQModel):
     layer_modules = [
         ["self_attn.k_proj", "self_attn.v_proj", "self_attn.q_proj"],
         ["self_attn.out_proj"],
-        ["fc1"],
-        ["fc2"],
+        # ["fc1"], disabled: not a good candidate for quantization
+        # ["fc2"], disabled: not a good candidate for quantization
     ]


LeiWang1999:
Looks like this option will cause a crash for vllm when load from opt-gptq model.

```bash
rank0]:     return loader.load_model(model_config=model_config,
[rank0]:   File "/root/vllm-bitblas/vllm/model_executor/model_loader/loader.py", line 270, in load_model
[rank0]:     model.load_weights(
[rank0]:   File "/root/vllm-bitblas/vllm/model_executor/models/opt.py", line 355, in load_weights
[rank0]:     param = params_dict[name]
[rank0]: KeyError: 'model.decoder.layers.0.fc1.weight'
```

Qubitium:
@LeiWang1999 Yes. there are some interface regressions (bitblas unittests) failing when post this PR merge. In current main, bitblas backend with OPT fc1/fc2 modeuls skipped is outputing non-sense output. But backend=exllamav2 is working fine. 

Do you have any idea why?

Btw, this PR is currently required since without bypass fc1/fc2, OPT will fail our pre/post quantization PPL regression test. From our tests, fc1/fc2 are not quantizable. avg_loss is astronomical at all N layers.

Qubitium:
Just to add, we tested AutoGPTQ main as well and the same avg_loss is observed there so this is not unique to our refractors. The OPT model quants likely never had optimal quantization from the very beginning and due to PPL tests not performed on pre/post quant, it was not discovered until now.

Qubitium:
UPDATE: @ZX-ModelCloud  still checking inference regression regarding this PR.  Either we need to revert this our PPL tests are broken not reflecting real-generate results. We will adjust accordingly.

PR #131, opened 2024-07-01
add gemma2 support

LRL-ModelCloud:


Qubitium:
Status update: testing passing for gemma 2-9B (non-it and it) but generation quality failed for 27B. Very verry strange. We will add to limit code execution to only 9B models for now.

Issue #132, opened 2024-07-01
[BUG] Cannot install importlib as it's a builtin package


LeiWang1999:
Hi all, I'm trying to build this project from source, however the build pipeline is falling at `pip install -r requirements`:

```bash
pip install importlib>=1.0.4
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [19 lines of output]
      Error processing line 1 of /opt/conda/lib/python3.10/site-packages/protobuf-3.20.2-py3.10-nspkg.pth:
      
        Traceback (most recent call last):
          File "/opt/conda/lib/python3.10/site.py", line 186, in addpackage
            exec(line)
          File "<string>", line 1, in <module>
        ModuleNotFoundError: No module named 'importlib.util'
      
      Remainder of file ignored
      Error processing line 1 of /opt/conda/lib/python3.10/site-packages/ruamel.yaml-0.17.21-py3.10-nspkg.pth:
      
        Traceback (most recent call last):
          File "/opt/conda/lib/python3.10/site.py", line 186, in addpackage
            exec(line)
          File "<string>", line 1, in <module>
        ModuleNotFoundError: No module named 'importlib.util'
      
      Remainder of file ignored
      ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

It seems that the importlib shouldn't be installed manually, as it's a built in lib from python 3.1.

ref: https://docs.python.org/3/library/importlib.html

PR #133, opened 2024-07-01
[Fix] Export TORCH_CUDA_ARCH_LIST in install.sh

LeiWang1999:
This pull request includes a change to the `install.sh` script to export the `TORCH_CUDA_ARCH_LIST`  variables in install.sh scripts. Otherwise, the build pipeline of cuda backend will crash.

Also remove `importlib` from requirements.txt ref to #132


Qubitium:
@LeiWang1999  Thanks! LGTM.  I removed PTX flag since we don't want JIT of the cuda codes for now. We observed slightly faster code with PTX disabled.

PR #135, opened 2024-07-01
Calculate ppl on gpu

ZYC-ModelCloud:


gptqmodel/utils/perplexity.py
```diff
@@ -192,11 +192,14 @@ def _process_batch(self, i, n_ctx, n_batch, tokens, nll, count):
 
         for j in range(min(512, n_ctx // 2), n_ctx - 1):
             tok_logits = logits[0][0][j].cpu().numpy()
+            # Convert to tensor and move to device
```
Qubitium: looks like reversal of line 194. Might as well just remove `.cpu().numpy()` away from line 194?

PR #137, opened 2024-07-01
[BACKEND] Add QBits support

CSY-ModelCloud:


gptqmodel/nn_modules/qlinear/qlinear_qbits.py
```diff
@@ -0,0 +1,322 @@
...
+    elif dtype == torch.bfloat16:
+        return "bf16"
+    elif isinstance(dtype, str) and dtype in ["int8", "fp32", "fp16", "bf16"]:
+        return dtype
+    else:
+        assert False, "Unsupported pytorch dtype {} to str dtype".format(dtype)
+
+
+class QBitsQuantLinear(nn.Module):
+    QUANT_TYPE = "qbits"
```
Qubitium: QUANT_TYPE has been removed in main.

gptqmodel/nn_modules/qlinear/qlinear_qbits.py
```diff
@@ -0,0 +1,322 @@
...
+        outfeatures,
+        bias,
+        kernel_switch_threshold=128,
+        trainable=False,
+        weight_dtype=torch.bfloat16,
+        **kwargs,
+    ):
+        super().__init__()
+
+        if bits not in [4, 8]:
```
Qubitium: use SUPPORTED_BITS = [ n , n ] cls property override instead of runtime code

gptqmodel/utils/model.py
```diff
@@ -404,12 +405,17 @@ def gptqmodel_post_init(model, use_act_order: bool, max_input_length: Optional[i
...
     model_uses_exllama = False
+    model_uses_qbits = False
+
     # exllamav2
     fixed_bytes = {}
     model_uses_exllamav2 = False
 
     for name, submodule in model.named_modules():
-        if isinstance(submodule, ExllamaQuantLinear):
+        if isinstance(submodule, QBitsQuantLinear) and submodule.QUANT_TYPE == "qbits":
```
Qubitium: second part of check is not needed.

Qubitium:
@CSY-ModelCloud There are two issues for Qbits that requires fixing before I can merge:

1. Qbits use dynamic sym. It checks for zero tensors and if condition are met, it switches sym=False to sym=True. This becomes a problem since we store the entire quant model's sym value as config.json. We need to detect this in `quantize()` double check if it is doing dynamic sym in the qunanztiation stage also. Return this value from method so we can correctly set the sym value in config.json

2. refractor `.asym`  to use our code's norm of `.sym`.  asym is opposite of sym so after rename we need to reverse the conditions.

gptqmodel/nn_modules/qlinear/qlinear_qbits.py
```diff
@@ -0,0 +1,331 @@
...
+    QBITS_EXCEPTION = e
+
+if QBITS_AVAILABLE:
+    from intel_extension_for_transformers import qbits  # with QBits kernels ()
+
+# TODO FIXME: qbits support 2, 3 bits also so dtype mapping should be fixed
+BITS_DTYPE_MAPPING = {
+    4: "int4_clip",
+    8: "int8",
+}
```
Qubitium, line 16: @PenghuiCheng  Qbits supports 2, 3 bits also but the dtype mapping is missing 2, 3 support. Can you tell us what is the correct qbits dtype mapping for 2, 3 bits?

Qubitium (approved):

Qubitium (approved):

Qubitium (approved):

Qubitium:
@PenghuiCheng We have merged Qbits support with unit tests in this PR and a few follow-up PRs.

PR #165, opened 2024-07-04
[REFRACTOR] Remove Backend.CUDA and Backend.CUDA_OLD

ZX-ModelCloud:


Qubitium:
Both cuda/cuda-old have no good working case in July 2024. They perform much worse than exllama and exllama v2 kernels. The only saving grace is more bits supported but 99.9% of the cases will be using 4bit. We will not spend the time to support compat for 4 kernels when we can just pick the fastest 2.

PR #183, opened 2024-07-08
[FIX] Marlin padding

LRL-ModelCloud:


gptqmodel/nn_modules/qlinear/qlinear_marlin.py
```diff
@@ -157,8 +176,15 @@ def pack(self, linear, scales):
                 self.bias = linear.bias.clone()
 
     def forward(self, A):
+        import torch.nn.functional as F
         A = A.half()
+        if A.size(-1) != self.infeatures and self.infeatures > A.size(-1):
+            A = F.pad(A, (0, self.infeatures - A.size(-1)))
```
Qubitium: A.size(-1) is called max 3 times if condition matches.

gptqmodel/nn_modules/qlinear/qlinear_marlin.py
```diff
@@ -157,8 +176,15 @@ def pack(self, linear, scales):
                 self.bias = linear.bias.clone()
 
     def forward(self, A):
+        import torch.nn.functional as F
```
Qubitium: Move this import to top of file.

PR #190, opened 2024-07-09
[CORE] Add vLLM Backend for FORMAT.GPTQ

PZS-ModelCloud:


gptqmodel/models/base.py
```diff
@@ -590,8 +591,13 @@ def forward(self, *args, **kwargs):
 
     def generate(self, **kwargs):
         """shortcut for model.generate"""
-        with torch.inference_mode(), torch.amp.autocast(device_type=self.device.type):
-            return self.model.generate(**kwargs)
+        load_format = kwargs.pop('load_format', None)
```
Qubitium: Remove `load_format`, there is no need to pass this. Set model._backend to `backend` and use it.

gptqmodel/models/base.py
```diff
@@ -1126,6 +1132,25 @@ def skip(*args, **kwargs):
             )
             load_checkpoint_in_model = True
             quantize_config.format = FORMAT.GPTQ_V2
+        if backend == BACKEND.VLLM:
+            if quantize_config.format != FORMAT.GPTQ and quantize_config.format != FORMAT.GPTQ_V2:
```
Qubitium: vllm cannot run with gptq_v2, only format.gptq.

gptqmodel/models/base.py
```diff
@@ -590,8 +591,13 @@ def forward(self, *args, **kwargs):
...
     def generate(self, **kwargs):
         """shortcut for model.generate"""
-        with torch.inference_mode(), torch.amp.autocast(device_type=self.device.type):
-            return self.model.generate(**kwargs)
+        load_format = kwargs.pop('load_format', None)
+        with torch.inference_mode():
+            if load_format == 'vllm':
+                return vllm_generate(self.model, **kwargs)
+            else:
+                torch.amp.autocast(device_type=self.device.type)
```
Qubitium, line 599: autocast is a context, not a statement. Need `with ...`

gptqmodel/models/base.py
```diff
@@ -1126,6 +1132,25 @@ def skip(*args, **kwargs):
...
+                raise ValueError(f"{backend} backend only supports FORMAT.GPTQ: actual = {quantize_config.format}")
+
+            if backend == BACKEND.VLLM:
+                model = load_model_by_vllm(
+                    model=model_name_or_path,
+                    trust_remote_code=trust_remote_code,
+                    **kwargs,
+                )
+                model.config = model.llm_engine.model_config
+                model.config.model_type = "vllm"
```
Qubitium: vllm requires  setting `model.config.model_type`? Why do we need to override `model.config`?

tests/test_vllm.py
```diff
@@ -0,0 +1,36 @@
...
+        "The capital of France is",
+        "The future of AI is",
+    ]
+    sampling_params = SamplingParams(temperature=0.8, top_p=0.95)
+    def test_load_vllm(self):
+        model = GPTQModel.from_quantized(
+            self.MODEL_ID,
+            device="cuda:0",
+            backend=BACKEND.VLLM,
+            trust_remote_code=True,
```
Qubitium: why do we need `trust_remove_code=True` for tinyllama?

gptqmodel/utils/vllm.py
```diff
@@ -0,0 +1,24 @@
...
+from vllm import LLM, SamplingParams
+
+def load_model_by_vllm(
+    model,
+    trust_remote_code,
+    **kwargs,
+):
+    model = LLM(
+        model=model,
+        trust_remote_code=trust_remote_code,
```
Qubitium: There is no need to explicitly add `trust_remove_code` here. it is passed, if needed, via **kargs

tests/test_vllm.py
```diff
@@ -0,0 +1,36 @@
...
+            gpu_memory_utilization=0.2
+        )
+        outputs = model.generate(
+            load_format="vllm",
```
Qubitium: Remove `load_format` arg.  Check if `model.config` is instance of vllm llm_engine.config since it was set in ```model.config = model.llm_engine.model_config```

gptqmodel/models/base.py
Qubitium: `autocast` is a context? The original code is using `with autocast` but here it is just a statement.

gptqmodel/models/base.py
Qubitium: Check later comment to see if we can infer type of model.config to vllm_config.

PR #259, opened 2024-07-22
fix issue 253

PZS-ModelCloud:


gptqmodel/models/base.py
```diff
@@ -618,20 +619,30 @@ def save_quantized(
 
         state_dict = model.state_dict()
 
+        to_removes = _remove_duplicate_names(state_dict)
```
Qubitium: @PZS-ModelCloud  What this block of new code for? Add comments.

Qubitium (approved):

PR #276, opened 2024-07-23
[REFRACTOR] Cleanup backend and model_type usage

LRL-ModelCloud:


Qubitium:
@LRL-ModelCloud  @ZX-ModelCloud  I can't merge this yet. This introduces new `backend` property to `BaseModel` but the `backend` prop is not correctly updated to reflect the actual value during runtime. 

For example, I load `from_quantized` model and set `backend=auto`, but since it is auto (not real), it will be converted to real backend after "auto" checking for compatible backend for this model. Once the backend is selected, the `backend=auto` value should reflect the real `backend` and no longer `auto`.

PR #300, opened 2024-07-25
add "gptqmodel_marlin_cuda_24" CUDAExtension

ZX-ModelCloud:


Qubitium:
@ZX-ModelCloud  There are two marlin kernels in vllm. One is dense (the one that's compatible with GPTQModel) and spare(2-4) which uses/loads special 2-4 sparsified models.

PR #302, opened 2024-07-26
[FIX] missing .py files for models using trust_remote

PZS-ModelCloud:


gptqmodel/models/base.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import os
 import re
+import shutil
```
Qubitium, line 6: Is this pkg part of python? If not , add to req.txt?

gptqmodel/models/base.py
```diff
@@ -80,6 +81,8 @@ def __init__(
         quantize_config: QuantizeConfig,
         qlinear_kernel: nn.Module = None,
         load_quantized_model: bool = False,
+        trust_remote_code: bool = False,
+        original_model_name_or_path: str = None,
```
Qubitium: Why do we need to pass `original_model_name_or_path`? Isn't this always just model_path_or_name?

gptqmodel/utils/model.py
```diff
@@ -633,3 +634,18 @@ def check_to_quantized(config):
         if config.bits > 8 or "fp" in config.data_type or "float" in config.data_type:
             return False
         return True
+
+def copy_remote_file(save_dir, file_extension=".py", model_id_or_path=""):
```
Qubitium: rename to more accurate`copy_py_files` since we actually don't have a way to detect what files are for `trust_remote_code`.

gptqmodel/utils/model.py
```diff
@@ -5,7 +5,8 @@
 import os
 from logging import getLogger
 from typing import List, Optional
-
+from huggingface_hub import HfApi, hf_hub_download
```
Qubitium, line 8: Another `hugginface_hub` depend that is not part of req.txt. Will crash for users that do have it installed.

tests/test_quant_unsupport_transformers.py
```diff
@@ -0,0 +1,52 @@
...
+        model.quantize(self.calibration_dataset, batch_size=64)
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            model.save_quantized(
+                tmp_dir,
+            )
+
+            del model
+            py_files = [f for f in os.listdir(tmp_dir) if f.endswith('.py')]
+            self.assertTrue(len(py_files) > 0, f"No .py files found in the directory: {tmp_dir}")
```
Qubitium: instead of checking for. > 0, match EXACTLY what .py files are copied.

Find a model that has remote_code for both tokenizer and model:

Some `trust_remote_code` models have .py files for both tokenizer and model inference.

gptqmodel/models/base.py
PZS-ModelCloud: shutil is part of python, no need add req

gptqmodel/models/base.py
PZS-ModelCloud: update to model_name_or_path
cls or quantize_config not set model_path_or_name, save_quantize can not get original_model_path_or_name

PR #303, opened 2024-07-26
[CORE] Use vllm/nm dense marlin kerenl

ZX-ModelCloud:


Qubitium:
@ZX-ModelCloud  need to update `validatae_marlin_compat` method so that `desc_act` is supported since this new kernels supports both 8bits and desc_act.

PR #338, opened 2024-08-08
add lm_eval api

PZS-ModelCloud:


Qubitium:
Need refractor:

The current pr is just shell for lm_eval. We need closer integration.

m = GPTQModel.from_quantized()
m.lm_eval()

PR #385, opened 2024-09-09
support for minicpm3

LDLINGLINGLING:
I am a staff member of Model Best. We have recently open sourced minicpm3.0. This is a model address that uses gtpq to quantify minicpm3.0. It has achieved good results. We look forward to your merger.

https://huggingface.co/openbmb/MiniCPM3-4B-GPTQ-Int4


Qubitium:
@LDLINGLINGLING  Thank you for the PR and great minicpm models! LGTM.

PR #386, opened 2024-09-13
Fix missing link

mrT23:
First of all, kudos for this project.
It's the only project I found that properly supports modern models like llama-3.1 out of the box. Also the speed, and other factors, seem better.

Fixing in this PR a small bug - the previous link led to a missing page.


Qubitium:
Thanks! LGTM

PR #401, opened 2024-09-26
[MODEL] Add mllama support

LRL-ModelCloud:


Qubitium:
Note: This PR is PARTIAL MLamma support. Only text layers (layeres repsonsible for text generation) are quantized. The Cross-attention layers are not yet quantized. As result, 4bit quant of 11B vision model results in 50% reduction and not ~1/4 as expected.

PR #535, opened 2024-11-06
[API]  Replace from_pretrained and from_quantized with unified load()

ZYC-ModelCloud:
Ad `load()` to unify `from_pretrained` and `from_quantized` .

Remove overrding/passing `quantize_config` in `from_quantized`.


gptqmodel/models/auto.py
```diff
@@ -129,6 +170,7 @@ def from_quantized(
         quantize_config: Optional[QuantizeConfig | Dict] = None,
         use_safetensors: bool = True,
         trust_remote_code: bool = False,
+        format: Optional[FORMAT] = None,
```
Qubitium: Why do we need to add `format` here?

gptqmodel/models/auto.py
```diff
@@ -102,6 +104,45 @@ def __init__(self):
...
+            model_name_or_path: Optional[str],
+            quantize_config: Optional[QuantizeConfig | Dict] = None,
+            device_map: Optional[Union[str, Dict[str, Union[str, int]]]] = None,
+            max_memory: Optional[dict] = None,
+            device: Optional[Union[str, int]] = None,
+            backend: BACKEND = BACKEND.AUTO,
+            use_safetensors: bool = True,
+            trust_remote_code: bool = False,
+            verify_hash: Optional[Union[str, List[str]]] = None,
+            format: Optional[FORMAT] = None,
```
Qubitium: why format? `format` is alreaedy passed as `quantize_config.format`

Qubitium (commented):
Reemove passing and overriding of quantize_config in load() stage. Quant config should always be loaded via saved model config.

I cannot think of a plausible scenario where a quant model is saved without config and needs an override. This legacy code from autogptq make is much complex than necessary.

PR #676, opened 2024-11-27
use temp dir for saving test quantized model

CSY-ModelCloud:


CSY-ModelCloud:
fixed 2+ tests running together, saved models will be overwriten

Qubitium (approved):

PR #706, opened 2024-11-29
fix cpu/xpu backend selection

jiqing-feng:
Still need this change to run `quantization/basic_usage.py` in a CPU-only device.


Qubitium:
https://github.com/ModelCloud/GPTQModel/pull/705/files

Waiting for XPU node access to verify fix. As well as check this PR on XPU

PR #736, opened 2024-12-04
add QUANT_TYPE in qlinear

jiqing-feng:
For optimum and transformers check qlinear quant type tests


Qubitium:
@jiqing-feng  Can you add a short commen above all QUANT_TYPE like

```
# for transformers/optimum compat
```

So we do accidentally refractor this unused var in the future.

PR #1106, opened 2025-01-20
[MODEL] Add Telechat2 (China Telecom)

1096125073:
Add support for the Telechat model (VLLM now supports the Telechat model).

### DIFF
--- a/gptqmodel/models/auto.py
+++ b/gptqmodel/models/auto.py
@@ -86,6 +86,7 @@ from .definitions.stablelmepoch import StableLMEpochGPTQ  # noqa: E402
 from .definitions.starcoder2 import Starcoder2GPTQ  # noqa: E402
 from .definitions.xverse import XverseGPTQ  # noqa: E402
 from .definitions.yi import YiGPTQ  # noqa: E402
+from .definitions.telechat2 import TeleChat2GPTQ
 
 logger = setup_logger()
 
@@ -139,6 +140,7 @@ MODEL_MAP = {
     "hymba": HymbaGPTQ,
     "olmo2": Olmo2GPTQ,
     "ovis": OvisGPTQ,
+    "telechat":TeleChat2GPTQ,
 }
 
 
--- a/gptqmodel/models/auto.py
+++ b/gptqmodel/models/auto.py
@@ -84,9 +84,10 @@ from .definitions.qwen2_vl import Qwen2VLGPTQ  # noqa: E402
 from .definitions.rw import RWGPTQ  # noqa: E402
 from .definitions.stablelmepoch import StableLMEpochGPTQ  # noqa: E402
 from .definitions.starcoder2 import Starcoder2GPTQ  # noqa: E402
+from .definitions.telechat2 import TeleChat2GPTQ
 from .definitions.xverse import XverseGPTQ  # noqa: E402
 from .definitions.yi import YiGPTQ  # noqa: E402
-from .definitions.telechat2 import TeleChat2GPTQ
+
 
 logger = setup_logger()
 
@@ -140,7 +141,7 @@ MODEL_MAP = {
     "hymba": HymbaGPTQ,
     "olmo2": Olmo2GPTQ,
     "ovis": OvisGPTQ,
-    "telechat":TeleChat2GPTQ,
+    "telechat": TeleChat2GPTQ,
 }
 
 
--- a/gptqmodel/models/definitions/telechat2.py
+++ b/gptqmodel/models/definitions/telechat2.py
@@ -1,9 +1,12 @@
 from ..base import BaseGPTQModel
+import torch
 
 
 class TeleChat2GPTQ(BaseGPTQModel):
     # telechat2 requires custom model code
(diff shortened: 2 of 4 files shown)


Qubitium:
@1096125073 Do you have HF model url for this model? We need to download this model and ci test

1096125073:
> @1096125073 Do you have HF model url for this model? We need to download this model and ci test

Yes, you can try this one
[https://huggingface.co/Tele-AI/TeleChat2-7B](url)

gptqmodel/models/auto.py
```diff
@@ -139,6 +140,7 @@
     "hymba": HymbaGPTQ,
     "olmo2": Olmo2GPTQ,
     "ovis": OvisGPTQ,
+    "telechat":TeleChat2GPTQForCausalLM,
```
Qubitium: Please rename to `TeleChat2GPTQ`.

gptqmodel/models/definitions/telechat2.py
```diff
@@ -0,0 +1,20 @@
+from ..base import BaseGPTQForCausalLM
+
+
+class TeleChat2GPTQForCausalLM(BaseGPTQForCausalLM):
```
Qubitium: 1. Rename `TeleChat2GPTQ`
2. You need to override base class prop `require_trust_remote_code` and set to True here since this model requires remote code loading.

gptqmodel/models/definitions/telechat2.py
```diff
@@ -0,0 +1,20 @@
...
+    layer_type = "TelechatBlock"
+    layers_block_name = "transformer.h"
+    outside_layer_modules = ["transformer.word_embeddings", "transformer.ln_f"]
+
+    """
+    If other frameworks are used for inference (such as VLLM),
```
Qubitium, line 16: Question. Is QKV fused inside the weights and that's why we should not quantize? We support qkv fused kernels in vLLM even if original weights are not fused.

Qubitium:
@1096125073 Thank you for the PR. I will have @LRL-ModelCloud  take over the PR changes from here. He will fix this PR so it can run correctly in GPTQModel with CI tests. Right now, this code will not run since gptqmodel uses different structure on how to define models.

1096125073:
> @1096125073 Thank you for the PR. I will have @LRL-ModelCloud take over the PR changes from here. He will fix this PR so it can run correctly in GPTQModel with CI tests. Right now, this code will not run since gptqmodel uses different structure on how to define models.

Sorry, I just discovered that the properties of this library are different from AutoGPTQ and have been updated

gptqmodel/models/definitions/telechat2.py
1096125073: This is because in the Telechat model, the weights of kV are distributed in an interleaved manner. If head_dim is 64 and quantized group_size is 128, it result in a pair of kV weights sharing the same scaler, which may lead to errors.

You can refer to the load_weights function in the following  link
[https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/telechat2.py](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/models/telechat2.py)

Qubitium:
> > @1096125073 Thank you for the PR. I will have @LRL-ModelCloud take over the PR changes from here. He will fix this PR so it can run correctly in GPTQModel with CI tests. Right now, this code will not run since gptqmodel uses different structure on how to define models.
> 
> Sorry, I just discovered that the properties of this library are different from AutoGPTQ and have been updated

No problem. We wil fix this. But stop force-pushing so we can fix this PR.

Qubitium:
@1096125073  We are testing. 

1. If you are with telechat2 team, please ask manager to add `torch_dtype` to `config.json`

https://huggingface.co/Tele-AI/TeleChat2-7B/blob/main/modeling_telechat2.py#L186

The assert codes here are strange. Why is it forcing `cuda`? GPTQModel supports inference on multiple devices and does not require cuda.

1096125073:
> @1096125073 We are testing.
> 
> 1. If you are with telechat2 team, please ask manager to add `torch_dtype` to `config.json`
> 
> https://huggingface.co/Tele-AI/TeleChat2-7B/blob/main/modeling_telechat2.py#L186
> 
> The assert codes here are strange. Why is it forcing (asserting) `cuda`? GPTQModel supports inference on multiple devices and does not require cuda.

Sorry, this is being handled by another team, but we are currently organizing the code and submitting the PR to the transformer library. I believe this will be resolved soon.

Qubitium:
@1096125073  We are running into `dtype` assert bugs in `forward` if we load the original model as `bfloat16`. `float16` can run. Can you confirm the model `.bin` files are native bfloat16 or float16?

1096125073:
> @1096125073 We are running into `dtype` assert bugs in `forward` if we load the original model as `bfloat16`. `float16` can run. Can you confirm the model `.bin` files are native bfloat16 or float16?

yes, 7b run as  float16

Qubitium (approved):

PR #1127, opened 2025-01-21
[MODEL] add deepseek v3 support

LRL-ModelCloud:


James4Ever0:
Does it supports dynamic 1.58bit quantization, as Unsloth does to Deepseek R1 671B?

PR #1128, opened 2025-01-21
cherry pick AutoGPTQ/pull/770. fix the issue of qlinear packing being too slow.

CSY-ModelCloud:
https://github.com/AutoGPTQ/AutoGPTQ/pull/770/


CSY-ModelCloud:
@Qubitium

Issue #1161, opened 2025-01-24
[BUG] Failed building on ROCm  - requires CUDA_VERSION environment variable


isaranto:
**Describe the bug**

I'm trying to build GPTQModel from source for ROCm 6.2 with the following command:

```
ROCM_VERSION=6.2 pip install -v . --no-build-isolation
```
and get the following error:
```
Trying to compile GPTQModel for CUDA, but Pytorch 2.5.1+rocm6.2 is installed without CUDA support.
```
Looking at the code in [setup.py](https://github.com/ModelCloud/GPTQModel/blob/main/setup.py#L102) there seems to be an explicit check for CUDA_VERSION and an error is thrown if it is not provided.
So it seems that if I just provide a dummy value for cuda version the package is built successfulyy as the `get_version_tag()` function will return the rocm version in [line 110](https://github.com/ModelCloud/GPTQModel/blob/main/setup.py#L110)

A simple fix would be to change the order of the commands in the get_version_tag function.
Alternatively we would be extracting the relevant information from torch as we do for CUDA.
For CUDA we check `torch.version.cuda` but this attribute is empty on pytorch/rocm builds while the attribute `torch.version.hip` would return something like '6.2.41133-dd7f95766'. We would need to extract the major+minor version from this string.
```

**Software Info**

Debian bookworm+ python 3.11 + Pytorch 2.5.1+rocm6.2

PR #1162, opened 2025-01-24
fix(build): fix building from source with ROCm

isaranto:
Fix for #1161 
Trying to build the package from source with ROCm for AMD GPUs fails as an environment variable `CUDA_VERSION` is required but not used anywhere.

Here I'm following the simple fix by changing the order of execution in the `get_version_tag` function in setup.py.
More information can be found in the Issue #1161

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,6 +98,9 @@ def get_version_tag() -> str:
 
     import torch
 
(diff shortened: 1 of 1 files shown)


Qubitium (approved):
LGTM!

PR #1184, opened 2025-01-30
Update/Refractor Bitblas/Marlin/Cuda

Qubitium:


gptqmodel/nn_modules/qlinear/bitblas.py
```diff
@@ -389,7 +360,7 @@ def forward(self, A):
         A_void = ctypes.c_void_p(A.data_ptr())
         # m is the product of the last n - 1 dimensions of A
         m = ctypes.c_int32(reduce(operator.mul, A.shape[:-1], 1))
-        self.bitblas_matmul.call_lib(
+        self.bitblas_matmul.lib.call(
             A_void , *self.q_params, ctypes.c_void_p(C.data_ptr()), m
```
Qubitium, line 364: @LeiWang1999 Happry Chinese New Years! When you have time, can you point to us or help us find the cause of bitblas v0.1.0 issue in this PR? Thank you. Looks like C api has changed and not sure how to debug this. 

We are getting `Segmentation fault (core dumped` at the `self.bitblas_matmul.lib.call(`. 0.1.0 no longer has `call_lib` but we see there is a `lib.call` but trying to use it just seg faults:

For context, this PR finally tries to update bitblas depend from very old 0.0.1-dev3 to latest 0.1.0, and also use the `repack_from_gptq_v2` code (copied) from bitblas repo so that we repack v2 weights after qweight load before bitblas `forward`. 

Please use the following code to replicate with this PR code:

```py
from gptqmodel import GPTQModel, BACKEND

model = GPTQModel.load(
    "ModelCloud/Qwen2.5-0.5B-Instruct-ci-test-bitblas",
    backend=BACKEND.BITBLAS, # change to BACKEND.TORCH or MARLIN for normal generate
)

print(model.tokenizer.decode(model.generate("What is the capital of United States?")[0]))


```

PR #1281, opened 2025-02-16
Fix supported models list (syntax error)

Forenche:
* There was a missing ',' after minicpm which was breaking model detection
* I noticed this after trying to use the PR: https://github.com/ModelCloud/GPTQModel/pull/1116

* Test: Build on local and try out minicpmo


Qubitium:
Thanks for the fix! merged.

PR #1283, opened 2025-02-17
feat: load model from modelscope

suluyana:
Hello,

First of all, thank you very much for your group's high-quality open-source LLM model compression/quantization toolkit. In this pull request, I have added an environment variable called "GPTQMODEL_USE_MODELSCOPE". With this environment variable, Chinese users can directly retrieve models from ModelScope, which is significantly faster.

ModelScope is the largest model community in China. It serves as a premier platform designed to bridge model checkpoints with model applications, offering the essential infrastructure to facilitate the sharing of open models and promote model-centric development. For more information, visit our GitHub page: [ModelScope](https://github.com/modelscope).


gptqmodel/__init__.py
```diff
@@ -19,3 +19,8 @@
 from .utils import BACKEND
 from .utils.exllama import exllama_set_max_input_length
 from .version import __version__
+
+import os
+if os.getenv('GPTQMODEL_USE_MODELSCOPE', 'False').lower() == 'true':
```
Qubitium: Please check for both `true` and `1` since many env vars have mixed true vs 1 usage.

tests/test_modelscope.py
```diff
@@ -0,0 +1,26 @@
...
+
+    @classmethod
+    def setUpClass(self):
+        self.MODEL_ID = "Qwen/Qwen2.5-0.5B-Instruct-GPTQ-Int4"
+
+    def test_load_modelscope(self):
+        model = GPTQModel.load(self.MODEL_ID)
+
+        result = model.generate("Hello")[0]
+        assert(model.tokenizer.decode(result))
```
Qubitium: Please use a better prompt + assert to make sure the model generates but also generates correctly.

Prompt: "The capital of mainland China is"
Assert that "beijing" or "bei-jing" is in the `result.lower()`

Qubitium (commented):
@suluyana  ModelScope is great for the China ai community. Thank you for the PR. 

Only some nitpicks.

gptqmodel/__init__.py
Qubitium: @suluyana  Should we use a generic `USE_MODELSCOPE` env var that is not tied to GPTQMODEL?

suluyana: > @suluyana Should we use a generic `USE_MODELSCOPE` env var that is not tied to GPTQMODEL?

The main concern is that modelscope also supports other projects. If different projects use the same generic env var USE_MODELSCOPE, it may not be flexibly supported in the user's environment.

Current env var follow the same naming rules as other projects:
image
image

Qubitium:
@suluyana  All good!

PR #1284, opened 2025-02-17
allow passing model instance to evalplus & update tokenizer loading logics

CSY-ModelCloud:


CSY-ModelCloud:
@Qubitium

Qubitium:
Broken auto `apply_chat_template` logic. Need to revert.

PR #1309, opened 2025-02-21
Eora Fused Kernel

ZX-ModelCloud:


Qubitium:
Note: This PR add EoRA fused kernel from Nvidia (@hutm) based on EoRA paper (@nbasyl). Kernel code is based on the vllm-exllama gptq kernel with fused EoRA operations added in.

Issue #1376, opened 2025-03-05
[BUG] deepseek r1 quantization AttributeError: 'GPTQ' object has no attribute 'H'


yuchiwang:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/caf3e99c-e1dc-4544-8c21-a9297a87db90)

when do quantization: ds r1 moe layer

**GPU Info**

Show output of:

![Image](https://github.com/user-attachments/assets/ef54c8bd-5d6d-49c1-b5d9-ccb22d3c4e44)

**Software Info**

Operation System/Version + Python Version

Show output of:
```
pip show gptqmodel torch transformers accelerate triton
```

![Image](https://github.com/user-attachments/assets/87d0c65e-da3e-482b-923f-dd4f11a504a5)

**If you are reporting an inference bug of a post-quantized model, please post the content of `config.json` and `quantize_config.json`.**

**To Reproduce**

How to reproduce this bug if possible.

![Image](https://github.com/user-attachments/assets/d1bad703-c8ec-46f4-8035-6f787a999394)

**Expected behavior**

**Model/Datasets**

Make sure your model/dataset is downloadable (on HF for example) so we can reproduce your issue.

**Screenshots**

**Additional context**

yuchiwang:
i want to konw how gpu memory need for ds3 or ds r1，There is no way to reduce gpu memory requirements？ thanks

yuchiwang:
I use cpu device， still this error happen. I have 2t cpu memory. 

![Image](https://github.com/user-attachments/assets/63ad78a9-2e2f-443f-aab1-42214a3f2a21)

yuchiwang:
![Image](https://github.com/user-attachments/assets/c72a11d8-91a0-4acd-a223-e1ef698d163a) 

![Image](https://github.com/user-attachments/assets/b91eff17-d925-4b2c-8d8b-f6f99c7ee618) 每次推理只激活部分专家

Qubitium:
> ![Image](https://github.com/user-attachments/assets/c72a11d8-91a0-4acd-a223-e1ef698d163a) 
> 
> ![Image](https://github.com/user-attachments/assets/b91eff17-d925-4b2c-8d8b-f6f99c7ee618) 
> 原因清楚了，每次moe推理只激活部分专家并，在进行quantize时需要跳过未激活的专家，否则因为这些专家没有add batch会导致error

You need to feed it more calibration data so all moe layers are hit.

[...]

PR #1389, opened 2025-03-07
add fwd counter

yuchiwang:
issue [1376 ](https://github.com/ModelCloud/GPTQModel/issues/1376).  Add a counter and print in the quant logs how many actual add_batch or inputs are captured by the quantization process for each module.


Qubitium:
@yuchiwang  Thanks for the PR. This will be very useful for MoE modules.

PR #1437, opened 2025-03-12
fix the error when downloading models using modelscope

mushenL:
Hi, I noticed an issue while using the gptqmodel. Even though I’ve set the environment variables properly, the model isn’t being downloaded from ModelScope. It seems like the code might be preventing the modelscope snapshot_download logic from executing entirely.


Qubitium:
@mushenL  Thank you! LGTM.

PR #1502, opened 2025-04-04
Multi GPU Quantization

Qubitium:
Prelim framework to allow quantization to operate on 2 gpu. Just set `CUDA_VISIBLE_DEVICES=` to two gpus and gptqmodel will use the second GPU for quantization phase. For now, only 2 GPU is used even if you provide > 2. 

However, due to the structure of the current code, peak vram saving is <15% (based on a simple testing)  and quantization speed takes a hit since we need to move multi-GB worth of tensors to/back from GPU0 <-> GPU1.  However 15% peak vram saving is a substantial if your model is just too large for the calibration dataset size on one gpu.

Make sure the two GPUs have at least PCIE 4.0 x4 lanes each.  

This is just the first phase of multi-gpu quantization .

### DIFF
--- a/examples/benchmark/perplexity.py
+++ b/examples/benchmark/perplexity.py
@@ -21,6 +21,7 @@ from gptqmodel.utils import Perplexity
 from transformers import AutoTokenizer
 
 os.environ["CUDA_DEVICE_ORDER"] = "PCI_BUS_ID"
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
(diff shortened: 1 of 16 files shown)


chesterout:
How do we use 2 gpus. I tried two ways, 
1. in the code: os.environ["CUDA_VISIBLE_DEVICES"] = "0,1".
2. in the linux command, export CUDA_VISIBLE_DEVICES=0,1. 
However, none of these works. Can you clarify?

PR #1508, opened 2025-04-06
llama 4

Qubitium:


Qubitium:
We are missing layers/modules. The quantized sized vs non-quantized size is only 6.2% which is not normal. At 4bit, it whould be closer to 70% difference.

Qubitium:
Need to fix this. MoE modules are parameters and not linear so they are skipped for quantization.


[mmexport1743945031722.png](https://github.com/user-attachments/assets/ac3b3fd4-50eb-420e-929f-c6ae1f2c8b11)

PR #1510, opened 2025-04-07
Switch to non-deprecated mlx.core.clear_cache()

smpanaro:
Removes some log noise: `mx.metal.clear_cache is deprecated and will be removed in a future version. Use mx.clear_cache instead.`


Qubitium:
@smpanaro  Thank you! LGTM.

PR #1512, opened 2025-04-07
Dream Model Support

Qubitium:
@jiacheng-ye We have never tested diffusion model quantization but can you check if this PR works? Thanks! I am not even sure what type of calibration dataset to use since difussion generate seems pretty funky, in a good way. =)

### DIFF
--- a/gptqmodel/models/auto.py
+++ b/gptqmodel/models/auto.py
@@ -20,6 +20,7 @@ import os
 
 import threadpoolctl
 
+from .definitions.dream import DreamGPTQ
 from ..utils.logger import setup_logger
 
 log = setup_logger()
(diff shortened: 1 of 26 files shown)


Qubitium:
ARC eval passing (bits=8, groups_size=128), 256 samples. 

```bash
--------Eval Result---------
|    Tasks    |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|-------------|------:|------|-----:|--------|---|-----:|---|-----:|
|arc_challenge|      1|none  |     0|acc     |↑  |0.3558|±  |0.0140|
|             |       |none  |     0|acc_norm|↑  |0.3695|±  |0.0141|
```

PR #1527, opened 2025-04-10
Replace module hooking with tree-defined targeting

Qubitium:
Even though existing code does not error, we should never ever hook/wrap a module that will never particpates  in quantization. This doesn't make any sense. The fix is much more complicated and likely require for 2-3 PRs to refractor. 

Currently add optional pin-point module targeting via `tree` of `base.layers_modules_tree`.  This will play nicely into more generic multi-modal support as mm models embeds up to 3 separate models essentially into 1. The existing static-non-tree based config will be harder to support. 

Tree syntax

```py
# Full tree of quantizable modules
# `#` str will match any number: useful for layers and moe indexing.
# List[str] for serial linked nodes. List str are linear depth linked modules presented in a linear fashion with no divergence.
# Dict{str: List[str] | Dict | Tuple[str]} for diverging nodes where a node splits into multiple paths/nodes.
# Tuple(str) for final targeted modules/nodes: there are only strings representing the final targeted modules
layers_modules_tree = [
    "model",
    "layers",
    "#",
    {
        "self_attn": ("k_proj", "v_proj", "q_proj", "o_proj"),
        "mlp": ("up_proj", "gate_proj", "down_proj"),
    }
]
 ```

### DIFF
--- a/gptqmodel/looper/eora_processor.py
+++ b/gptqmodel/looper/eora_processor.py
@@ -29,7 +29,7 @@ from ..models import BaseGPTQModel
 from ..models.writer import (PROCESS_LOG_FWD_TIME, PROCESS_LOG_LAYER, PROCESS_LOG_MODULE,
                              PROCESS_LOG_NAME, PROCESS_LOG_TIME, PROCESS_MAX_MEMORY)
 from ..quantization.config import QuantizeConfig
-from ..quantization.gptq import CPU
+from ..quantization.gptq import CPU, CUDA_0, CUDA_1
 from ..utils.logger import setup_logger
 from ..utils.model import move_to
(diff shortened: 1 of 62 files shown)


Qubitium:
@Cecilwang While testing for conv1d/2d support I found deep issues/oversight with the current hooking targeting mechnaism so this Pr kind of spiraled out of control. But the code relevant to you is minimal. You just need to add the `layer_modules_tree` definition to your model for pinpoint control of hooks and check if the `hooked_linear.py` changes is usable for your model. I could not find a model that has nn.conv1d or nn.conv2d that is actually quantized. Let me know if you know any, other than mamba which I don't think we support yet.

Issue #1529, opened 2025-04-10
a strange way to slove the nan loss during quantization


ywlq:
I'm trying to quantize the Qwen2-57B-A14B model using the following configuration:

```
pretrained_model_id = "/mnt/82_store/xj/modelzoo/qwen/Qwen/Qwen2-57B-A14B"
quantized_model_id = "/mnt/82_store/xj/GPTQModel/quantmodel/Qwen2-57B-A14B-2bit-128g-c4

 calibration_dataset = load_dataset(
     "allenai/c4",
     data_files="en/c4-train.00001-of-01024.json.gz",
     split="train"
   ).select(range(1024))["text"]

quantize_config = QuantizeConfig(
    bits=2,
    group_size=128,
    dynamic={
         r".*mlp\.shared_expert.*": { "bits": 8, "group_size": 128 },
         r".*mlp\.experts.*down_proj.*": { "bits": 8, "group_size": 128 },
     }
)

model = GPTQModel.load(pretrained_model_id, quantize_config)
model.quantize(calibration_dataset, auto_gc=False, batch_size=4)
model.save(quantized_model_id)
```

it dont work :(

However, the quantization fails with NaN errors. After tracing the issue, I found that it originates from this function during weight quantization:

```
def quantize(x, scale, zero, maxq, requires_groupwise_processing: bool):
    if maxq < 0:
        return (x > scale / 2).float() * scale + (x < zero / 2).float() * zero
    if requires_groupwise_processing:
        q = torch.clamp(torch.round(x / scale), -maxq, maxq)
        return scale * q
    else:
        q = torch.clamp(torch.round(x / scale) + zero, 0, maxq)
        return scale * (q - zero)
```
The problem occurs when scale == 0, which leads to division by zero and causes NaN. I temporarily fixed it with this patch:

```
def quantize(x, scale, zero, maxq, requires_groupwise_processing: bool):
    if maxq < 0:
        return (x > scale / 2).float() * scale + (x < zero / 2).float() * zero
    if requires_groupwise_processing:
        scale = torch.where(scale == 0, torch.tensor(1e-8, device=scale.device), scale)  # 防止除以零
        q = torch.clamp(torch.round(x / scale), -maxq, maxq)
        return scale * q
    else:
        scale = torch.where(scale == 0, torch.tensor(1e-8, device=scale.device), scale)  # 防止除以零
        q = torch.clamp(torch.round(x / scale) + zero, 0, maxq)
[...]

PR #1531, opened 2025-04-10
Sidestep NaN caused by div by zero where calculated scale == 0

ywlq:
Sidestep NaN caused by div by zero where calculated scale == 0. The issue address is [1529](https://github.com/ModelCloud/GPTQModel/issues/1529)

### DIFF
--- a/gptqmodel/quantization/quantizer.py
+++ b/gptqmodel/quantization/quantizer.py
@@ -30,11 +30,11 @@ def quantize(x, scale, zero, maxq, requires_groupwise_processing: bool):
     if maxq < 0:
         return (x > scale / 2).float() * scale + (x < zero / 2).float() * zero
     if requires_groupwise_processing:
-        scale = torch.where(scale == 0, torch.tensor(1e-8, device=scale.device), scale)
+        scale = torch.where(scale == 0, torch.tensor(1e-8, device=scale.device), scale) # avoid division by zero
         q = torch.clamp(torch.round(x / scale), -maxq, maxq)
         return scale * q
     else:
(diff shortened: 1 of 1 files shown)


gptqmodel/quantization/quantizer.py
```diff
@@ -30,9 +30,11 @@ def quantize(x, scale, zero, maxq, requires_groupwise_processing: bool):
     if maxq < 0:
         return (x > scale / 2).float() * scale + (x < zero / 2).float() * zero
     if requires_groupwise_processing:
+        scale = torch.where(scale == 0, torch.tensor(1e-8, device=scale.device), scale)
```
Qubitium: Just some small change requests:

1. Move the scale mod to outside of the if/else.
2. Wrap with check before mod and log so we know this module/model has problems and not just totally ignore this issue.:

```py
if (scale == 0).any():
   scale = torch.where(scale == 0, torch.tensor(1e-8, device=scale.device), scale)
   log.error(f"Quantize: zero scale found and bypassed. This module may not be properly trained or received enough calibration data causing unmanageable outliers. If MoE module, please check with model maker and/or feed with larger calibration data. : `{scale}`. ")

Qubitium (changes requested):
@ywlq Looks good. just some nitpicks.

gptqmodel/quantization/quantizer.py
Qubitium: @ywlq  I don't see any code changes? Please make the small the changes above and push.

Qubitium (changes requested):
Check the small changes requested above.

PR #1550, opened 2025-04-19
Fix rotation for tied embedding models

smpanaro:
When embeddings are tied, fusing the `pre_head_layernorm` into `lm_head` also impacts `embed_tokens`. This happens in a few common models: Qwen2.5-0.5B, Llama3.2 for example. Possible this is the source of the issue mentioned in https://github.com/ModelCloud/GPTQModel/pull/1425#issuecomment-2717194335.

The solution is to untie the weights.

Also made some related changes along the way, happy to revert them or split them into separate PRs:
- Allow rotation for all quant_methods. May not always help, but doesn't need to be QQQ-specific.
- Couple small fixes to improve MPS compatibility. (Not perfect, e.g. still need to comment out `rotate_ov_proj`.)

### DIFF
--- a/gptqmodel/models/base.py
+++ b/gptqmodel/models/base.py
@@ -421,7 +421,7 @@ class BaseGPTQModel(nn.Module):
                                     f"current model is {self.__class__.__name__}")
 
             if self.model.config.tie_word_embeddings:
-                print("Rotation requires word embeddings to be untied. Untying.")
+                log.info("Rotation requires word embeddings to be untied. Untying.")
                 self.model.config.tie_word_embeddings = False
                 lm_head, _ = get_module_by_name_prefix(self.model, self.lm_head)
                 lm_head.weight = nn.Parameter(lm_head.weight.data.clone())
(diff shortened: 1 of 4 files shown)


gptqmodel/models/base.py
```diff
@@ -412,30 +412,40 @@ def quantize(
...
+        # rotate model
+        if self.quantize_config.rotation:
+            from gptqmodel.models.definitions.llama import LlamaGPTQ
+            from gptqmodel.models.definitions.qwen2 import Qwen2GPTQ
+            if not isinstance(self, (LlamaGPTQ, Qwen2GPTQ)):
+                raise ValueError(f"rotation only supports: llama/qwen2 model, "
+                                    f"current model is {self.__class__.__name__}")
+
+            if self.model.config.tie_word_embeddings:
+                print("Rotation requires word embeddings to be untied. Untying.")
```
Qubitium: @smpanaro Can you replace with `print` with `log.info` here for consistency since all `prints` should be `logs` now.

Qubitium (approved):
@smpanaro Thank you for the PR, I see you also fixed internal api compat with QQQ in addition to the rotation fix as well as mps torch.compile compat. Nice!!!

Just a note on `print` to `log.info`.

gptqmodel/models/base.py
smpanaro: No problem, done!

PR #1554, opened 2025-04-21
Fix input processing for convolution

Cecilwang:
Thanks for the previous support for convolution layers. In this PR, I’ve added fixes for two specific cases:

- Conv1D and unfold parameter requirement:
The unfold operation expects 2D parameters, so for Conv1D layers, both the input tensor and parameters need to be expanded by one dimension to match the expected shape.

- Grouped convolutions:
Fixed the input tensor shape for convolutions that use the groups parameter to ensure the input is correctly formatted and grouped during computation.

ref:
https://pytorch.org/docs/stable/generated/torch.nn.Unfold.html
https://discuss.pytorch.org/t/conv1d-implementation-using-torch-nn-functional-unfold/109643/3

### DIFF
--- a/gptqmodel/quantization/gptq.py
+++ b/gptqmodel/quantization/gptq.py
@@ -242,14 +242,36 @@ class GPTQ:
         if isinstance(self.module, (nn.Linear, transformers.Conv1D)):
             reshaped_inp = reshaped_inp.reshape(-1, reshaped_inp.shape[-1])
         else:
-            unfold = nn.Unfold(
-                self.module.kernel_size,
-                dilation=self.module.dilation,
-                padding=self.module.padding,
-                stride=self.module.stride,
-            )
-            # output size (batch_size, channels * \prod kernel_size, num_patches)
-            reshaped_inp = unfold(reshaped_inp)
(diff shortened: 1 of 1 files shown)


Qubitium:
@Cecilwang Do you know of a opensource model currently on HF that would trigger this code? I want want to add `test_conv1d.py` tests to CI so we can check for regressions in the future. I know you are currently applying to a private model but was wondering if you know a non-private model that we can do a simple quant/inference on?

Otherwise, this PR is great and ready to go!

Cecilwang:
`state-spaces/mamba-130m-hf` is a [MambaForCausalLM](https://github.com/huggingface/transformers/blob/fee1190601b5d04ec6d3f7f58fd22788d7f3236d/src/transformers/models/mamba/modeling_mamba.py#L274) whose `slow_forward` function might trigger this PR's logical. You can try it without installing `causal_conv1d`. 

https://github.com/huggingface/transformers/blob/main/src/transformers/models/mamba/modeling_mamba.py#L54-L61
https://github.com/huggingface/transformers/blob/main/src/transformers/models/mamba/modeling_mamba.py#L326-L328

PR #1559, opened 2025-04-25
Supprt add glm4

glide-the:
Reproduction：https://huggingface.co/glide-the/GLM-4-9B-Chat-0414-identity-4bits/


Qubitium:
@glide-the Thank you!

PR #1616, opened 2025-05-20
Add Falcon H1 model support

HamzaYousLM:
This pull request adds support for the Falcon H1 model to the GPTQModel repository. The following changes have been made:

- Added the falcon_h1 definition to the `gptqmodel/models/definitions` folder.
- Imported the falcon_h1 model in `gptqmodel/models/definitions/__init__.py`.
- Included falcon_h1 in the MODEL_MAP within `gptqmodel/models/auto.py`.

These updates enable the integration and usage of the Falcon H1 model within the framework.

### DIFF
--- a/gptqmodel/models/definitions/falcon_h1.py
+++ b/gptqmodel/models/definitions/falcon_h1.py
@@ -25,8 +25,7 @@ class FalconH1GPTQ(BaseGPTQModel):
     layers_node = "model.layers"
     layer_type = "FalconH1DecoderLayer"
     layer_modules = [
-        ["feed_forward.gate_proj"],
-        ["feed_forward.up_proj"],
+        ["feed_forward.gate_proj", "feed_forward.up_proj"],
         ["feed_forward.down_proj"],
     ]
 
--- a/gptqmodel/models/definitions/falcon_h1.py
+++ b/gptqmodel/models/definitions/falcon_h1.py
@@ -1,0 +1,33 @@
+# Copyright 2024-2025 ModelCloud.ai
+# Copyright 2024-2025 qubitium@modelcloud.ai
+# Contact: qubitium@modelcloud.ai, x.com/qubitium
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ..base import BaseGPTQModel
+
+
+class FalconH1GPTQ(BaseGPTQModel):
+
+    base_modules = ["model.embed_tokens", "model.norm_f"]
+    pre_lm_head_norm_module = "model.norm_f"
+
+    layers_node = "model.layers"
+    layer_type = "FalconH1DecoderLayer"
+    layer_modules = [
+        ["feed_forward.gate_proj"],
+        ["feed_forward.up_proj"],
+        ["feed_forward.down_proj"],
+    ]
+
+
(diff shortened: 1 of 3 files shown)


Qubitium:
@HamzaYousLM Thanks for the PR! I just a few quick questions. 

1. is falcon h1 released to the public? I want to check on the modeling.py file to see if the correct order of layer modules is set. 
2.  related to 1,  right now `gate` `up` and `down` are separate, usually `gate` and `up` are together [`gate`, `up`] and `down` calculations depends on the result of both gate and up where gate and up do not depend on each other. 

Asking because if `down` calculations is dependendent on result of `gate` + `up` and `gate` and `up` are not dependedent in the modeling forward, we can make the quantization faster and more accurate by grouping `gate` and `up` together. This depends on the forwarding code of the modeling.py file for falcon h1.

gptqmodel/models/definitions/falcon_h1.py
```diff
@@ -0,0 +1,33 @@
...
+    base_modules = ["model.embed_tokens", "model.norm_f"]
+    pre_lm_head_norm_module = "model.norm_f"
+
+    layers_node = "model.layers"
+    layer_type = "FalconH1DecoderLayer"
+    layer_modules = [
+        ["feed_forward.gate_proj"],
+        ["feed_forward.up_proj"],
+        ["feed_forward.down_proj"],
+    ]
```
Qubitium: @HamzaYousLM  If falcon h1 is like falcon e which is llama based, the layer modules should be following for faster quantization. Pleaes check. 

```py
layer_modules = [
        ["feed_forward.gate_proj", "feed_forward.up_proj"],
        ["feed_forward.down_proj"],
    ]
```

HamzaYousLM:
> @HamzaYousLM Thanks for the PR! I just a few quick questions.
> 
> 1. is falcon h1 released to the public? I want to check on the modeling.py file to see if the correct order of layer modules is set.
> 2. related to 1,  right now `gate` `up` and `down` are all separate, usually `gate` and `up` are together [`gate`, `up`] and `down` calculations depends on the result of both gate and up where gate and up do not depend on each other.
> 
> Asking because if `down` calculations is dependendent on result of `gate` + `up` and `gate` and `up` are not dependedent in the modeling forward, we can make the quantization faster and more accurate by grouping `gate` and `up` together. This depends on the forwarding code of the modeling.py file for falcon h1.
> 
> For most transformer models:
> 
> ```
> [ gate, up ],
> [ down ]
> ```

HamzaYousLM:
Hi @Qubitium, 

Thank you for your review, 

1. Yes, it is publicly available, here is the link to the modeling file https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon_h1/modeling_falcon_h1.py. 
2. It is a hybrid architecture, and the feed forward part is similar the the llama architecture.

gptqmodel/models/definitions/falcon_h1.py
HamzaYousLM: I have applied the requested change in the latest commit.

Qubitium:
> Hi @Qubitium,
> 
> Thank you for your review,
> 
> 1. Yes, it is publicly available, here is the link to the modeling file https://github.com/huggingface/transformers/blob/main/src/transformers/models/falcon_h1/modeling_falcon_h1.py.
> 2. It is a hybrid architecture, and the feed forward part is similar the the llama architecture.

@HamzaYousLM 
1. Please add copy `https://github.com/ModelCloud/GPTQModel/blob/main/tests/models/test_falcon.py` code and add a `falcon_h1.py` ci test so we can protect this model from future regressions in CI tests. 
2. Also check that `FalconH1Attention` is quantizable. Right now, the quant model def file does not include the qkvo layers which based my check on the modeling file should be quantizable as it is based on Llama? 

https://github.com/huggingface/transformers/blob/163138a911c1fb4451ec4b32edaee20918a59def/src/transformers/models/falcon_h1/modeling_falcon_h1.py#L333C7-L333C24

For llama style attention, `[q, k, v], [o]` is usually the layout and the two subsets goes above gate/up+down in priority as the attention modules are exectued first before the feedforward layers.

Qubitium:
@HamzaYousLM  We have ported your code into different PR and tested simple quant + inference and it works. Please do more testing as well. Thanks. Closing this PR. 

https://github.com/ModelCloud/GPTQModel/pull/1621

PR #1627, opened 2025-06-15
feat(gemma3): also support larger gemma3 models and not only small te…

joennlae:
This PR add support for Gemma3 models larger than 1B. They have a vision encoder and are wrapped inside the `Gemma3ForConditionalGeneration`.


Qubitium:
@joennlae Thank you for the PR!

PR #1673, opened 2025-08-09
lm_head hooked

Chunfei-He:
While testing with \GPTQModel\tests\test_lm_head.py, I encountered an error where the lm_head layer was not being quantized:
ERROR `lm_head` was not invoked, if it is a MoE module, it may lack sufficient calibration data routed to it.
image
which eventually caused the process to raise this error.Traceback (most recent call last):
ValueError: No quantization for lm_head module
image

The root cause is that when using **replace_module_with_hooked_tree**, the lm_head is not hooked
image

To address this, I added a separate piece of code to explicitly hook the lm_head, ensuring that it is properly included in the quantization process.


Qubitium:
@Chunfei-He Thank you for the bug fix! Code looks clean and it is indeed an oversight. Great catch and fix.

PR #1675, opened 2025-08-11
fixed bugs when quantize lm_head

528-dev:
The issue with quantizing the language model head (lm_head) has been described in [ModelCloud/GPTQModel#1673 ](https://github.com/ModelCloud/GPTQModel/pull/1673). However, the proposed fix merely replaces the Linear layer with a HookedLinear, which is insufficient.

The key point is whether a LayerNorm layer exists before the lm_head. While inter-decoder-layer data propagation typically does not involve additional LayerNorms—hence no special handling in the original code—this assumption does not hold when quantizing the lm_head. If a LayerNorm precedes the lm_head, it must be properly identified and integrated into the quantization pipeline. Therefore, this scenario should not be overlooked during lm_head quantization.

### DIFF
--- a/gptqmodel/looper/module_looper.py
+++ b/gptqmodel/looper/module_looper.py
@@ -30,7 +30,7 @@ from ..looper.named_module import NamedModule
 from ..looper.native_processor import NativeProcessor
 from ..models import BaseGPTQModel
 from ..models._const import SUPPORTS_MODULE_TYPES
-from ..nn_modules.hooked_linear import replace_module_with_hooked_legacy, replace_module_with_hooked_tree
+from ..nn_modules.hooked_linear import replace_module_with_hooked_legacy, replace_module_with_hooked_tree, HookedLinear
 from ..utils.logger import setup_logger
 from ..utils.model import (find_modules, get_device, get_module, get_module_by_name_prefix,
                            get_moe_layer_modules, move_to, nested_move_to)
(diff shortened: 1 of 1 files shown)


Qubitium:
@528-dev  Oops. I did not catch this PR before merging #1673. Can you sync with `main` and apply your additional logic which round out the corner cases as you have described? I will merge this asap once this PR has synced with `main`. Thanks.

Liujh528:
> @528-dev Oops. I did not catch this PR before merging #1673. Can you sync with `main` and apply your additional logic which round out the corner cases as you have described? I will merge this asap once this PR has synced with `main`. Thanks.

When quantizing the lm head, line 237 does not work, causing the norm before the lm head to be ignored. https://github.com/ModelCloud/GPTQModel/blob/70d88c544e654c1994220414ceed70ee408384c6/gptqmodel/looper/module_looper.py#L237

Qubitium:
@528-dev So this PR is still not enough to  resolve the corner case issue where a norm layer preceeds lm-head? Do you have a sample open source model that we can run ci tests on to verify and check the sequence of model layer execution/walk. Thanks.

Liujh528:
> @528-dev So this PR is still not enough to resolve the corner case issue where a norm layer preceeds lm-head? Do you have a sample open source model that we can run ci tests on to verify and check the sequence of model layer execution/walk. Thanks.

Yes, my PR solves this problem. I tested Qwen/Qwen2-0.5B-Instruct.

Qubitium:
@528-dev There is a merge conflict in the current PR. Please resolve this and I will mergr the fix. Thanks.

Qubitium (approved):

Qubitium:
@528-dev  LGTM! Thank you.

PR #1683, opened 2025-08-16
Add gpt-neo model definition

smpanaro:
Small model without QKV bias and without RoPE which is an unusual combination.

Tested on MPS:
```sh
❯ PYTORCH_ENABLE_MPS_FALLBACK=1 python examples/quantization/basic_usage_wikitext2.py
[...]
test is a good idea, but I'm not sure that it's a good idea.
I'm not
Perplexity: 38.3164 [400 of 400] ████████████████████████████████████████████████████████████████████████████████| 0:02:13 / 0:02:13 [400/400] 100.0%
Quantized Model EleutherAI/gpt-neo-125m-gptq avg PPL is 39.33393188434296
```


Qubitium:
@smpanaro  Thanks for the new model support. LGTM!

PR #1684, opened 2025-08-16
Skip compile if MPS and < torch 2.8.0

smpanaro:
Seems that MPS support was added in [2.8.0](https://github.com/pytorch/pytorch/issues/157957). 

Before that version (here on 2.6.0), you get an error:
```
torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
LoweringException: TypeError: 'NoneType' object is not callable
  target: aten.mean.default
  args[0]: TensorBox(StorageBox(
    Pointwise(
      'mps',
      torch.float32,
      def inner_fn(index):
          i0 = index
          tmp0 = ops.load(arg1_1, 769 * i0)
          return tmp0
      ,
      ranges=[768],
      origin_node=clone,
      origins=OrderedSet([clone])
    )
  ))

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information


You can suppress this exception and fall back to eager by setting:
    import torch._dynamo
    torch._dynamo.config.suppress_errors = True
```

Copied the log warning from:

https://github.com/ModelCloud/GPTQModel/blob/70d88c544e654c1994220414ceed70ee408384c6/gptqmodel/models/base.py#L1287-L1290

### DIFF
--- a/gptqmodel/utils/torch.py
+++ b/gptqmodel/utils/torch.py
@@ -25,6 +25,8 @@ from torch.cpu import StreamContext
 from ..utils.logger import setup_logger
 from . import has_gil, log_gil_required
 
+PYTORCH_MIN_VERSION_WITH_MPS_COMPILE = Version('2.8.0')
+
 HAS_CUDA = False
 HAS_XPU = False
 HAS_MPS = False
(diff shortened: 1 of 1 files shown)


Qubitium:
@smpanaro  LGTM! Curious, does this (torch.compile with torch >= 2.8) speed up your inference on M chips?

PR #1686, opened 2025-08-19
some models require explicitly specifying the value of use_cache

LRL2-ModelCloud:


Qubitium:
Notes: Some models, likely broken logic, does not check for model config `use_cache` during inference if the arg is not passed. Normally, model.config.use_cache should be used as default for model inference unless overriden during inference.

Issue #1665, opened 2025-07-30
[BUG] qwen3 quantization error and how to fix：RuntimeError: The size of tensor a (32) must match the size of tensor b


Panda-eat-meat:
# Error
Quantizing layer 1 of 35              [1 of 35] | 0:00:15 / 0:04:30 [2/36] 5.6%Traceback (most recent call last):
  File ".../project/quant/gptq/gptq_model_int4.py", line 59, in <module>
    model.quantize(
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/gptqmodel/models/base.py", line 450, in quantize
    return module_looper.loop(
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/gptqmodel/looper/module_looper.py", line 326, in loop
    module(*layer_input) if is_lm_head_module else module(*layer_input,
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/transformers/modeling_layers.py", line 94, in __call__
    return super().__call__(*args, **kwargs)
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1762, in _call_impl
    return forward_call(*args, **kwargs)
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/transformers/models/qwen3/modeling_qwen3.py", line 265, in forward
    hidden_states, _ = self.self_attn(
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1751, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File ".../anaconda/anaconda3/envs/gptq_model/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1762, in _call_impl
    return forward_call(*args, **kwargs)
[...]

PR #1687, opened 2025-08-19
[FIX] transformers compat

LRL2-ModelCloud:


Qubitium:
Fixes https://github.com/ModelCloud/GPTQModel/issues/1665

PR #1689, opened 2025-08-19
Update requirements.txt

LRL2-ModelCloud:


Qubitium:
Will merge after come CI tests passes. Worried about `numpy` upgrade causing regressions.

Qubitium (approved):

PR #1730, opened 2025-08-24
GLM-4.5-Air support

avtc:
And probably other glm4_moe models.

I have quantized with this definition the GLM-4.5-Air model with padding to support tp8 and it works in vllm with tp8.


Qubitium:
@avtc  LTGM and great work! I really appreciate you adding comments (documentation) to the layers def. This will help future model defs.

PR #1731, opened 2025-08-24
Add a config option: mock_quantization to simplify heavy computations…

avtc:
I have used it to check if quantized model GLM-4.5-Air could load with vllm/sglang.
With `mock_quantization` the time required to produce a 8 bit quant was about 2.5 hours + 45 minutes for packing.
While without `mock_quantization` the expected time to produce a 8 bit quant was about 9.5 hours + 45 minutes for packing.

The resulting model still produce meaningful responses.

Here are some comparison for Qwen3-0.6B model:
```
# original
auto-round --model="/home/ubuntu/.cache/huggingface/hub/models--Qwen--Qwen3-0.6B/snapshots/c1899de289a04d12100db370d81485cdf75e47ca" --eval --device 0 --tasks lambada_openai --eval_bs 16
|    Tasks     |Version|Filter|n-shot|  Metric  |   | Value |   |Stderr|
|--------------|------:|------|-----:|----------|---|------:|---|-----:|
|lambada_openai|      1|none  |     0|acc       |↑  | 0.3996|±  |0.0068|
|              |       |none  |     0|perplexity|↓  |24.7832|±  |1.0502|

# 8bit GPTQ 10 samples - without mock_quantization
auto-round --model="/home/ubuntu/models/Qwen3-0.6B-gptqmodel-w8g128" --eval --device 0 --tasks lambada_openai --eval_bs 16
|    Tasks     |Version|Filter|n-shot|  Metric  |   | Value |   |Stderr|
|--------------|------:|------|-----:|----------|---|------:|---|-----:|
|lambada_openai|      1|none  |     0|acc       |↑  | 0.3992|±  |0.0068|
|              |       |none  |     0|perplexity|↓  |24.7743|±  |1.0468|

# 8bit GPTQ 10 samples - with mock_quantization
auto-round --model="/home/ubuntu/models/Qwen3-0.6B-gptqmodel-w8g128-mock" --eval --device 0 --tasks lambada_openai --eval_bs 16
|    Tasks     |Version|Filter|n-shot|  Metric  |   | Value |   |Stderr|
|--------------|------:|------|-----:|----------|---|------:|---|-----:|
|lambada_openai|      1|none  |     0|acc       |↑  | 0.3972|±  |0.0068|
|              |       |none  |     0|perplexity|↓  |25.0656|±  |1.0616|
```

### DIFF
--- a/gptqmodel/models/_const.py
+++ b/gptqmodel/models/_const.py
@@ -42,7 +42,7 @@ class DEVICE(str, Enum):
     CPU = "cpu"  # All CPU: Optimized for IPEX is CPU has AVX, AVX512, AMX, or XMX instructions
     CUDA = "cuda"  # Nvidia GPU: Optimized for Ampere+
     XPU = "xpu"  # Intel GPU: Datacenter Max + Arc
-    MPS = "mps"  # MacOS GPU: Apple Silion/Metal)
+    MPS = "mps"  # MacOS GPU: Apple Silicon/Metal)
     ROCM = "rocm"  # AMD GPU: ROCm maps to fake cuda
 
     @classmethod
--- a/gptqmodel/models/base.py
+++ b/gptqmodel/models/base.py
@@ -46,15 +46,13 @@ from ..utils.importer import select_quant_linear
 from ..utils.logger import setup_logger
 from ..utils.model import (MODALITY, check_to_quantized, find_modules, get_device, get_module,
                            get_module_by_name_prefix, get_moe_layer_modules, move_to, nested_move_to, pack_model)
-from ..utils.torch import torch_compile, torch_empty_cache
+from ..utils.torch import torch_compile, torch_empty_cache, TORCH_HAS_COMPILE
 from ._const import CALIBRATION_DATASET_CONCAT_CHAR, CPU, DEFAULT_MAX_SHARD_SIZE, DEVICE, SUPPORTS_MODULE_TYPES
 from .loader import ModelLoader
 from .writer import (PROCESS_LOG_FWD_TIME, PROCESS_LOG_LAYER, PROCESS_LOG_MODULE, PROCESS_LOG_TIME,
                      QUANT_LOG_DAMP, QUANT_LOG_LOSS, QUANT_LOG_NSAMPLES, ModelWriter)
 
-# pytorch 2.6.0 fixes many compilation errors
-TORCH_MIN_VERSION_STR = '2.6.0'
-PYTORCH_MIN_VERSION_WITH_COMPILE = Version(TORCH_MIN_VERSION_STR)
+
 
 def check_support_param_buffer_assignment(*args, **kwargs):
     return False
@@ -1284,9 +1282,9 @@ class BaseGPTQModel(nn.Module):
             log.warn("model is not quantized, skip compiling...")
             return self
 
-        if Version(torch.__version__) < PYTORCH_MIN_VERSION_WITH_COMPILE:
+        if TORCH_HAS_COMPILE:
             self.compiled = False
-            log.warn(f"To use compile(), you need to have torch version >= {TORCH_MIN_VERSION_STR}, please "
+            log.warn(f"To use compile(), you need to have torch version >= 2.6.0, please "
                            f"upgrade it by `pip install -U torch torchaudio torchvision`")
             return self
 
--- a/gptqmodel/nn_modules/qlinear/torch_fused.py
+++ b/gptqmodel/nn_modules/qlinear/torch_fused.py
@@ -26,8 +26,26 @@ from ...models._const import DEVICE, PLATFORM
 from ...nn_modules.qlinear import BaseQuantLinear, PackableQuantLinear
 from ...utils.backend import BACKEND
 from ...utils.logger import setup_logger
+from ...utils.torch import TORCH_HAS_XPU_FUSED_OPS
 
 log = setup_logger()
+
+# TODO: not yet working for cuda/cpu fused int4 ops
+# def pack_scales_and_zeros(scales, zeros):
+#     assert scales.shape == zeros.shape
+#     # assert scales.dtype == torch.bfloat16
+#     # assert zeros.dtype == torch.bfloat16
+#     return (
+#         torch.cat(
+#             [
+#                 scales.reshape(scales.size(0), scales.size(1), 1),
+#                 zeros.reshape(zeros.size(0), zeros.size(1), 1),
+#             ],
+#             2,
+#         )
+#         .transpose(0, 1)
+#         .contiguous()
+#     )
 
 class TorchFusedQuantLinear(PackableQuantLinear):
     SUPPORTS_BITS = [4]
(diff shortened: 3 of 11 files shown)


Qubitium:
> The resulting model still produce meaningful responses.

@avtc This is very surprising tbh. I wholeheartedly agreee this feature will help/accelerate with model testing since we can have a qunatized model, regardless of quality, for final eval pipeline testing well before the actual quant can be ready which can be 20+ hours for the super large models. The inference optimization/compat can start well before the actual quant is finished. 

> I have used it to check if quantized model GLM-4.5-Air could load with vllm/sglang. With `mock_quantization` the time required to produce a 8 bit quant was about 2.5 hours + 45 minutes for packing. While without `mock_quantization` the expected time to produce a 8 bit quant was about 9.5 hours + 45 minutes for packing.

Which CPU/GPU combo are you using for testing? 45 minutes for packing seems excessive for such as a small 0.6B model. 

The only minor issue I have is the `mock_quantization` property name. I will still merge this first, but we will likely rename it to `mock_hessian_inverse` to be more task specific in a future minor refractor PR.  The current code LTGM and very clean.

avtc:
> Which CPU/GPU combo are you using for testing? 45 minutes for packing seems excessive for such as a small 0.6B model.

I am using EPYC Genoa 9124 16 Core + 8 x 3090.
GLM-4.5-Air is 106B MoE model.
I was not able o run eval on it (autoround eval get an error), so provided evals from 0.6B model.

avtc:
The most time-consuming part is a loop `for i in range(count):`, while `_mock_hessian_inverse` does not recover much time and original `hessian_inverse` can be used. I am thinking about adding to config separate params:
```
    mock_hessian_inverse: bool = field(default=False, metadata={"help": "Use simplified hessian inverse (identity matrix)"})
    fast_loop: bool = field(default=False, metadata={"help": "Optimized version of quantization loop"})
```

Qubitium:
@avtc  You may want to try `python 3.13t` where `t` means `threading` or `GIL disabled` python. I know the naming is super confusing but 3.13t vs 3.13 means Python does not have global lock on threads making them REAL (unsafe) threads. 

I was able to 100x (or something ridiculous as we can now truely parallel cpu based packing on each core) the speed of packing and  get close to Nx the speed of quantization of MoE modules when using `gptqmodel` with `python 3.13t`. I have not tested lately since my free-threading code pushed to `main` was in april/may but you can try it. Just assign CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 when call ing gptqmodel and make sure you have an pyenv of `3.13t`. You should get a huge speedboost. Packing should be <60s as it will also use a lot of your cpu cores to do packing.

avtc:
> ```
>     mock_hessian_inverse: bool = field(default=False, metadata={"help": "Use simplified hessian inverse (identity matrix)"})
>     fast_loop: bool = field(default=False, metadata={"help": "Optimized version of quantization loop"})
> ```

I will do some tests and add another PR for that

avtc:
@Qubitium 
Also to note, when hessian_inverse is mocked with identity matrix, even the dataset of 10 samples works for GLM-4.5-Air, all expert modules quantized.
While without it, with fast_loop and 1024 samples - several experts on different layers were not quantized (leading to failed loading in vllm).
I will try without fast_loop later on.

---
I got a good speed boost with 3.13t for quantization, from 2.5hour to 0.5hours with 10 samples and fast_loop. While packing 45m -> 37m estimated, maybe because I have only 64Gb DDR5 RAM and use huge swap on nvme as a workaround of small RAM available.

Qubitium:
> I got a good speed boost with 3.13t for quantization, from 2.5hour to 0.5hours with 10 samples and fast_loop. While packing 45m -> 37m estimated, maybe because I have only 64Gb DDR5 RAM and use huge swap on nvme as a workaround of small RAM available.

Since you had 8 gpu a 2.5h to 0.5hour decrease is pretty good and on-par with my estimate. I think without nvme swap, it can go to 0.3 hours.  But 45m -> 37m for packing is much higher than my estimated improvement. And I do agree your 64GB ram is the limitating factor since the nvme takes a double hit as the packing needs contiguous ram which thrashes the nvme, while after packing it needs to write to nvme which also kills iops so nvme swap is likely causing at lest 2x speed decrease in packing stage. My estimate for without nvme swap is 45m -> <5m provided nvme is gen3 or higher.

avtc:
> I was able to 100x (or something ridiculous as we can now truely parallel cpu based packing on each core) the speed of packing and get close to Nx the speed of quantization of MoE modules when using `gptqmodel` with `python 3.13t`. I have not tested lately since my free-threading code pushed to `main` was in april/may but you can try it. Just assign CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 when call ing gptqmodel and make sure you have an pyenv of `3.13t`. You should get a huge speedboost. Packing should be <60s as it will also use a lot of your cpu cores to do packing.

Appeared I missed to do `export PYTHON_GIL=0`, as regex requirement forces it to `1` by default. Now the packing speed is much better.

Qubitium:
> Appeared I missed to do `export PYTHON_GIL=0`, as regex requirement forces it to `1` by default. Now the packing speed is much better.

How much faster now for packing that you have fully disabled GIL?

avtc:
> How much faster now for packing that you have fully disabled GIL?

GLM-4.5-Air (106B A12B)
1. samples: 10, batch_size: 1, mock_quantization: True, buffered_fwd: True
With `export PYTHON_GIL=0`, quantization time: 17m, packing time: 13m, then saving time 16m.
With `export PYTHON_GIL=1`, quantization time: 1h13m, packing time: 58m, then saving time 16m.

2. samples: 1024, batch_size: 1, mock_quantization: False, buffered_fwd: False
With `export PYTHON_GIL=0`, quantization time: 3h5m, packing time: 10m39s, then saving time 16m.

3. samples: 1024, batch_size: 1, mock_quantization: True, buffered_fwd: False
With `export PYTHON_GIL=0`, quantization time: 2h28m, packing time: 11m42s, then saving time 16m.

system: 8 x 3090 gen4 x8, 64Gb ddr5 4800, 768Gb swap on 4Tb 990Pro nvme (gen4), Epyc 9124

Qubitium:
@avtc  Btw, I saw your are working on cpu memory optizmations which is awesome. There is an actually memory leak within transformers/accelerate https://github.com/huggingface/transformers/issues/34366 that leaks mmaped memory. The problem is transformer/accelerate has no fine-grained support for releasing torch low level mmaped memory when loading models. So during quantization, we do layer by layer transformation, model memory should actually go down after each layer is processed. But transformer/accelerate loads the model as mmmap chunks via pytorch low level api but even if one chunk is still in use, it will hold on to to all mmaped memory for all previous calls. Transformers doesn't know or have no control which which mmap chunk to release. 

Not sure this issue still exists but I think it does. There was an internal PR where we reduced memory by `reloading` the model after each layer post quantization to force unloading the memory as a hack. This hack has not made it into the `main` yet and it is quite ugly. 

If we can fix this, we also fix the main cpu memory usage issue with the model itself. This is usually an issue with huge models like DeepSeek where the cpu memory usage approaches 2TB.

PR #1764, opened 2025-09-05
add build toggles for all cpp kernels

Qubitium:
@nbasyl


nbasyl:
Looks good to me! Will update you once the kernel is ready on our side. Thanks!

PR #1774, opened 2025-09-09
fix Q.to on multi-gpu gptq when proceeding fast and has many experts and gpus

avtc:
This fixes `Q.to` on multi-gpu gptq when proceeding fast and has many experts and gpus (for example with mock_quantization=True on 8 gpus on GLM-4.5-Air)

The error originally thrown:
`torch.AcceleratorError: CUDA error: invalid argument`
_For me it throws after first layer with experts finished processing._

Retry of `Q.to` fixes it.

Original code `Q.type_as` not only changes the type but also moves the tensor to `weight.data.device`
The logic of moving to device preserved.
And it seems that `wq = wq.to(device=DEVICE_0, non_blocking=False)` is redundant so was removed.

### DIFF
--- a/gptqmodel/looper/gptq_processor.py
+++ b/gptqmodel/looper/gptq_processor.py
@@ -30,7 +30,7 @@ from ..quantization import GPTQ, GPTQv2
 from ..quantization.config import QUANT_METHOD, QuantizeConfig
 from ..utils.logger import setup_logger
 from ..utils.model import move_to, pack_model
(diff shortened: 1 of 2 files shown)


Qubitium:
@avtc  Good catch! Looks like you found a torch/cuda sync bug ,or, gtqmode is not safely locking the cuda ops in our GIL=0 setup. I have a feeling it's the latter. Merge for now and log this for future TODO.

Issue #1680, opened 2025-08-14
[BUG] Some lays weight randomly don't quantize for  Qwen3-Coder-30B-A3B-Instruct


sunzj:
**Describe the bug**

when try to quantize Qwen3-Coder-30B-A3B-Instruct,  some lays don't quantize.  Quantized model.safetensors.index.json indicate following lays didn't do quantize.

    "model.layers.25.mlp.experts.100.down_proj.weight": "model-00003-of-00005.safetensors",
    "model.layers.25.mlp.experts.100.gate_proj.weight": "model-00003-of-00005.safetensors",
    "model.layers.25.mlp.experts.100.up_proj.weight": "model-00003-of-00005.safetensors",

about 0.4% weights were skipped.

**GPU Info**

Show output of:

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.120                Driver Version: 550.120        CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3090        On  |   00000000:4B:00.0 Off |                  N/A |
| 30%   28C    P8             35W /  350W |      18MiB /  24576MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA A100-PCIE-40GB          On  |   00000000:B1:00.0 Off |                    0 |
| N/A   32C    P0             32W /  250W |   37943MiB /  40960MiB |     19%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA GeForce RTX 3090        On  |   00000000:CA:00.0 Off |                  N/A |
| 30%   27C    P8             28W /  350W |       4MiB /  24576MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

```

**Software Info**

Operation System/Version + Python Version

[...]

PR #1775, opened 2025-09-10
Add fail safe

LRL-ModelCloud:


Qubitium:
Partially resolves #1680. The core issue is still modeling bug and/or the model has very uneven moe training distribution which cannot be fixed post-training.

avtc:
@Qubitium Can you please check the flow.

I am using GLM-4.5-Air with mock_quantization=False, and 1 sample.

When `fail_safe=False` there are such lines in log: 
`[31mERROR [0m mlp.experts.23.up_proj was not invoked, if it is a MoE module, it may lack sufficient calibration data routed to it.`

When `fail_safe=True`:
`[33mWARN [0m  Quantization: mlp.experts.23.up_proj is not activated due to model inference logic (MoE)`

But do not see lines containing:
`-> using fail safe mode. Please check if calibration data is sufficient.`

I assume this is expected as first pass (without fail_safe) produced warning, the second pass (with fail_safe) succeeded so the warning with `using fail safe mode` is absent.

Qubitium:
@avtc Will check tmr.

Qubitium:
@avtc  Should be fixed with https://github.com/ModelCloud/GPTQModel/pull/1780

Previous logic missed if/else branch.

PR #1778, opened 2025-09-10
TODO: Use threading lock to protect unsafe tensor moves in multi-gpu

Qubitium:


Qubitium:
@LRL-ModelCloud  Pleasea also test this PR asap and I will revert if there are issues.

avtc:
@Qubitium it works OK

PR #1780, opened 2025-09-10
fix multi gpu fail_safe

LRL-ModelCloud:


Qubitium:
fix https://github.com/ModelCloud/GPTQModel/pull/1775#issuecomment-3274104922

avtc:
This works.

Anyway consider throw an exception instead of log.error when fail_safe is False suggesting to use diversified inputs or enable fail_safe. Because when run quantization without saving log into the file, on the model having many experts and layers it's impossible to notice there were some experts unquantized, until trying to load into vllm and analyzing produced weights.

Or as alternative - the "dynamic" section of config must be adjusted with skipped modules, so vllm will load them unquantized. Idk if this will work OK, need to test.

PR #1807, opened 2025-09-17
hard deprecated ipex in favor of torch_fused

Qubitium:
Intel has deprecated and stopped development of IPEX related work and  XPU/CPU optimized kernel inference is moved to TorchFused with aten.ops from pytorch >= 2.8

Hard deprecate IPEX kernel and related code/format testing. Migrate the inference/quant test of IPEX to TorchFused instead.  

@jiqing-feng


jiqing-feng:
LGTM. I will also enable the torch fused linear for Intel CPU in the future.

PR #1809, opened 2025-09-17
Turtle in a half shell

Qubitium:
Ugly ugly hacks. Do not read this PR for your own sanity. It will be cleaned up later.  Lots of debug, random var names, etc. Seriously, don't read the commits. 

Current progress:

* 23.9% cpu memory saving for the entire end-to-end quantization + packing process. 
* 27% cpu memory saving if we only count the quantization stage without packing. Packing introduce more cpu ram usage which can be fixed in future commits. 

Target: > 75%


Qubitium:
6.4 GB to 4.1 GB quant ( excluding packing ) => 35.9% saving...

Pizza time!

Qubitium:
6.4 GB to 2.4 GB ( excluding packing ) => 62.5% saving...

Ice cream?

Qubitium:
6.4 GB to 1.7 GB ( excluding packing ) => 73.5% saving...

PR #1893, opened 2025-09-25
half fix

Qubitium:


mengxiaosen-patsnap:
Do you have plan working on it ? the bug V2 is still here!

PR #1930, opened 2025-09-26
fix inference mode not applied to threads

Qubitium:
@avtc  please test this which should fix your latest crash about cannot modif tensors inplace without inference mode ctx.


avtc:
yep, that fixed runtime error

PR #1932, opened 2025-09-26
fix eora compat

Qubitium:
@nbasyl  There is two issues with EoRA and main (lots of refractor and multi-threading changes in prep for pending multi-gpu data parallel inference). You can track this here. I will merge once `test_quant_and_eora` passes ci test. 

* module offloading to disk broke `eora` processor compat
* multi-gpu startup will cause `eora` to fail/crash.

### DIFF
--- a/gptqmodel/looper/eora_processor.py
+++ b/gptqmodel/looper/eora_processor.py
@@ -140,6 +140,8 @@ class EoraProcessor(LoopProcessor):
             device=module.target_device,
         )
 
+        del eigen_scaling_diag_matrix
+
         # wq with A/B applied
         computed_wq = wq + (B @ A)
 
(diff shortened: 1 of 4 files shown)


Qubitium:
Execution bug/order fixed due to threading scope issues. Offloading bug also fixed. But there is a secondary offloadig bug that appears to breaking lora module save. I will check this a bit later.

Qubitium:
Found the save bug. Will fix it soon.  The eora needs to join the offload disk life cycle. Right now main enjoys 75%+ cpu memory saving during quant with inline packing and offloading to disk right after processing. eora is missing this bit of logic so need to fix this.

Qubitium:
merging. ci runs but the quality is bad (not just eora). But other ci tests don't have this quality issue. not sure what's going on but looks to be this ci test specifically.

PR #1934, opened 2025-09-26
add eora toggle to ci test

Qubitium:


Qubitium:
@avtc  Ignore the PR title. Add a very estimated memory dealloc check and counter so we only call torch.cuda.empty_cache when we detect about ~1/4 of your gpu max vram has been deallocated recently. The default value is "auto" set in `memory.py` at the bottom. Right auto is set to min gpu vram / 4. You can set this if auto doesn't work for you. 

Goal is to avoid calling it as much as possible since it is very very slow. 

pass DEBUG=1 env before you call gptqmodel to get the memory dealloc/alloc debug logs.

PR #1939, opened 2025-09-27
Qwen3 omni moe support

LRL2-ModelCloud:


Qubitium:
Refractor: Add new `BaseQModel` properties

```py
ATTENTION_MASKS_DTYPE = torch.bool # default to bool
ATTENTION_MASKS_REQUIRED_FOR_INPUT : bool = False  # default to false
```

Qubitium:
@LRL2-ModelCloud 

```
 def pre_quantize_generate_hook_end(self):
        self.model.thinker.model.embed_tokens = self.model.thinker.model.embed_tokens.to(CPU)
        self.model.thinker.visual = self.model.thinker.visual.to(CPU)
        self.model.thinker.audio_tower = self.model.thinker.audio_tower.to(CPU)

        self.model.thinker.visual.rotary_pos_emb = self.model.thinker.visual.rotary_pos_emb.to(CPU)
        self.model.thinker.model.rotary_emb = self.model.thinker.model.rotary_emb.to(CPU)
```

Actually, call `offload_to_disk()` on these 3 modules so they directly go to disk and not waste cpu mem.

Qubitium:
@LRL2-ModelCloud 

add

```
INPUT_EMBEDDING_EXTRA_ARGS = None
```

This is for models like qwen vl whre we need to pass `return_audio=False` or `return_video=False` like args. We pass it as **INPUT_EMBEDDING_EXTRA_ARGS if not None.

PR #1941, opened 2025-09-28
use ext memlord pkg

Qubitium:
@avtc  Please test this. On llama 3.2 single GPU the peak vram reduced from consistent flat 4.3GB usage to 4.0GB with much more observed memory reclamation.  I expect the lower peak + much more frequent (high -> low transitions) will help with the MoE issue.  We now track memory usage directly via `pytorch` hooks and removed manual mem tracking from gptqmodel. Device level vram and usage is also monitored at 200ms intervals. 

There may be still be threading bugs (already fixed a few) so if you get segmentation fault, please provide the stack trace. 

This PR requires the following (unreleased memlord pkg)

 
```
pip install git+https://github.com/modelcloud/memlord
```

### DIFF
--- a/gptqmodel/looper/module_looper.py
+++ b/gptqmodel/looper/module_looper.py
@@ -48,7 +48,7 @@ class ModuleLooper():
         # auto gc mem profiler/tracker
         self.mem = MemLord()
         self.mem.hook_into_torch()
-        self.mem.hook_into_python()
+        #self.mem.hook_into_python()
         # self.mem.hook_into_python(
         #     enable_factory_wrappers=True,  # wrap common torch factories
         #     enable_deep_autowrap=True,  # also wrap many tensor methods + torch.* funcs
--- a/tests/models/test_qwen3_moe.py
+++ b/tests/models/test_qwen3_moe.py
@@ -14,6 +14,13 @@ class TestQwen3Moe(ModelTest):
     TRUST_REMOTE_CODE = True
     APPLY_CHAT_TEMPLATE = True
     EVAL_BATCH_SIZE = 6
+    V2 = False
+    DEBUG = True
+    ACT_GROUP_AWARE = True
+    DESC_ACT = False
+    DATASET_SIZE = 1024
+    DATASET_SORT = "desc"
+    QUANT_BATCH_SIZE = 4
 
     def test_mimo(self):
         self.quant_lm_eval()


Qubitium:
Confirmed there are stilll threading bugs. It will crash on 3.13t. Do not test yet.

Qubitium:
@avtc  Ready for testing. Validated under Llma 3.2 1B model. Right now the gc interval is set in module_looper's init class where it will init a MemLord instance that hooks into torch and python. There are still many bugs in counting (especially allocations) but at least it is gathering data in a threaded GIL=0 env without issue and no major bugs with python finalizers yet).  Please also pull latest from `memlord`: `pip install git+https://github.com/modelcloud/memlord`

The default `MemLord` autogc strategy is:

```py
    def set_auto_gc_strategy(self, strategy: Optional[dict] = None) -> None:
        if strategy is None:
            strategy = {
                (0, 50):   {"metric": "max", "threshold": {"percent": 15.0}},
                (50, 75):  {"metric": "max", "threshold": {"percent": 10.0}},
                (75, 101): {"metric": "max", "threshold": {"percent": 7.0,}},
            }
  ```

When the device (gpu, any gpu:index) is at vram utilization of 0-50 percent band (range), then it would apply the rule of `max` (wither counted +allocated or +freed, which ever is higher) meets the threshold which can be a fixed `bytes` or here `percent` of device vram capacity. 

There are still bugs in memlord where +allocation and +freed are still not very accurate. Right now, we have a semi-accurate estimate for +freed but very inaccurate reading of +allocated. Still, it is enough to give us values to use. 

You can play with the strategy to increase/decrease the jit auto-gc control. You can switch to `bytes` instead of percentage for more fixed states.

MemLord will print some stats when auto-gc is triggered. You can set `DEBUG=1` to enable verbose logs (way too much).

Qubitium:
Test failed (deadlock) for Qwen3_MoE. There is an multri-threaded issue where `torch.cuda.empty_cache()` cannot be called on any thread if there are pending gpu ops. It may corrupt the system causing asserts of deadlock the system. This worked in our old auto_gc code and if your patch is because the auto_gc is happneing on the main dispatcher thread that is dispatching thread to other gpus so at those point-in-time, it is not possilbe for any gpu to be operating (also protected by `async.join`.

avtc:
@Qubitium my patch is not about `auto_gc`, appeared patch works without `auto_gc`. Patch is about moving `ASYNC_WORKER.join()` from after the layers loop to the end of per-layer iteration. To wait for offload to finish before proceeding to next layer.

See: https://github.com/avtc/GPTQModel/commit/18175aaa0f080ce467ded2c0e4704a7dc2803564

Also to note, using the `ASYNC_WORKER.join()` behaviour works much much faster than `SERIAL_BG_QUEUE.join()`

--- update: I have not tested yet with SAMPLES > 1, checking right now how it will behave with 1024 samples.

PR #1961, opened 2025-10-01
Replicate + Turtle state fix

Qubitium:


Qubitium:
@avtc  so to confirm this PR fixed your `meta` load issues right? But still needed the `Q.to()` device lock to avoid the cuda memory errors. 

Btw, this PR also sped up MoE quantization by 5% or so.

avtc:
> @avtc so to confirm this PR fixed your `meta` load issues right? But still needed the `Q.to()` device lock to avoid the cuda memory errors.

Yep, fixed, Yep still need lock `Q.to()`. Tried with `CUDA_LAUNCH_BLOCKING=1` without lock `Q.to()` but deadlocked.

PR #1962, opened 2025-10-01
fix ctx for convert v1-v2 v2-v1 using in_place tensor mutation

Qubitium:


avtc:
Works!

Qubitium:
> Works!

Still need to find out why your Q.to() is failing. I even setup a cI test and it couldn't replicate the bug. It doesen't make sense. The Q.to() bug I mean.

avtc:
Have you tried with GLM-4.5-Air, mock_quantization=True, samples: `1`, 5+ gpus? On 3.13.7t?

PR #1963, opened 2025-10-01
Fix q to

Qubitium:
@avtc  Refractored code. May resolve your Q.to() crash.

### DIFF
--- a/gptqmodel/looper/module_looper.py
+++ b/gptqmodel/looper/module_looper.py
@@ -461,6 +461,18 @@ class ModuleLooper():
         position_ids = []
         layer_input_kwargs = []
 
+        try:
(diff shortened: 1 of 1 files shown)


avtc:
Confirmed, this pr fixed issue with `Q.to()`

Qubitium:
> Confirmed, this pr fixed issue with `Q.to()`

Finally! A throne in my backside finally plucked. You can pull `main` for a gptq quantization quality fix. You can now properly increase `batch_size` to as high as your gpu can manage to increase throughput at the expensive of more vram usage.

Each gpu/model pair has a sweet spot toggle for `batch_size`. There is no one size fits all. More batch sometimes is slower but not always.

PR #1966, opened 2025-10-02
Block h update

Qubitium:


Qubitium:
memory usage actually went up despite all my efforts..shelving this for now

PR #2000, opened 2025-10-10
enable cpu torch fused op

jiqing-feng:


Qubitium:
LGTM! Passing on both Intel and AMD CPU.

PR #2032, opened 2025-10-15
fix  attn mask

LRL2-ModelCloud:


Qubitium:
replace by https://github.com/ModelCloud/GPTQModel/pull/2034

PR #2033, opened 2025-10-15
Eora fix

Qubitium:
@nbasyl There is a huge bug within GPTQ + EoRa processing in this PR. The `samples` processed by gptq + eora forward hooks are not matching the tokens telemetry I have injected into the latest code. This may be regression or lay over from old code. Anyways. I will try to fix this soon. The PR will make EoRA run, it doesn't error but like said, the `samples` to tokens telemetry is widely out of sync.


Qubitium:
@nbasyl  Bugs fixed. Both single and multipe gpu acceleration for gptq + eora confirmed working via `test_quant_and_eora.py` ci test.

PR #2109, opened 2025-10-24
Fix GLM 4.5/4.6 and AIr not saving mtp layer after save (HF bug)

LRL2-ModelCloud:


Qubitium:
@avtc  Watch when this PR gets merged. I still need fix some details. This fixes a HF Transformers bug where if you save the model is MISSING a `mtp` layer for GLM Moe (4.5, 4.6 Air included) which is `optional` but used by vllm/sglang to speed up inference!

PR #2136, opened 2025-10-29
[MODEL] Add Qwen3-VL

techshoww:


Qubitium:
@techshoww  Thank you for the model contribution! I will merge this in the next 12 hours. I need run it against the real model for verification before merge.

Qubitium:
There is currently ci fail causing:

```py
RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cuda:0, different from other tensors on cpu (when checking argument in method wrapper_CUDA_bmm)
```

Once we find/fix this, this PR will be merged. Should be today.

Qubitium:
@techshoww  Merged. Thank you!

PR #2144, opened 2025-10-30
Fix triton version detection in nogil patcher

amd-vlarakic:
**Issue:**
Running GPTQModel based inference workload with **ROCm** generated **Triton** results in **ValueError** when converting patch version from string to integer:
```
File "/home/taccuser/workspace/dart/tmp7kf9o3i3_pyt_rocm_deepseek-r1-distill-llama-8b-gptq_inf/environment/lib64/python3.12/site-packages/gptqmodel/utils/nogil_patcher.py", line 42, in <genexpr>
    if version is None or tuple(int(part) for part in version.split(".")[:3]) < (3, 5, 0):
                                ^^^^^^^^^
ValueError: invalid literal for int() with base 10: '0+rocm7'
```

**Resolution:**
Fix applies more robust version comparison using **_packaging.version.Version_** module extensively used across the **GPTQModel** code base.

### DIFF
--- a/gptqmodel/utils/nogil_patcher.py
+++ b/gptqmodel/utils/nogil_patcher.py
@@ -9,9 +9,13 @@ import threading
 import time
 
 from .safe import ThreadSafe
+from importlib.metadata import version
+from packaging.version import Version, InvalidVersion
 
 
 _PATCHED_ATTR = "_gptqmodel_locked_save_file"
+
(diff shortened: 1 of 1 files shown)


amd-vlarakic:
Hi @Qubitium ,
Could you take a look?

Qubitium:
@amd-vlarakic  LGTM! Use `Version` is definitely better than the previous string split code. Thanks for the fix.

Issue #2116, opened 2025-10-25
When all modules of a layer excluded there is an error in forward replay


avtc:
@Qubitium

https://github.com/ModelCloud/GPTQModel/blob/e8a9fabc7e645a3b4d43d194fd3894cbc665dc5f/gptqmodel/looper/module_looper.py#L1324C1-L1332C33

```
                for index, names in enumerate(modules):
                    subset = self.crate_named_modules(full=full, is_lm_head_module=is_lm_head_module,
                                                      layer_index=layer_index, layers_prefix=layers_prefix,
                                                      names=names,
                                                      processor=processor,
                                                      fail_safe=fail_safe)

                    if len(subset) == 0:
                        continue
```
The variables: `subset_total`, `forward_device_map`, `subset_forward_serial`, `preserve_devices` are initialized inside of this loop (but after the `continue`), but after loop finished they are referenced in forward_replay.
And when subset len is 0 - they are not initialized.
Subset len is 0 when dynamic config exclude all modules of a layer.
For example for GLM-4.5-Air, dynamic config leave only moe modules to quantize:
```
dynamic = {
    r"-:model.embed_tokens.weight": {},
    r"-:.*shared_experts": {},
    r"-:.*shared_head": {},
    r"-:lm_head.weight": {},
    r"-:.*mlp.down": {},
    r"-:.*mlp.gate": {},
    r"-:.*mlp.up": {},
    r"-:.*post_attention_layernorm": {},
    r"-:.*self_attn": {},
    r"-:.*norm.weight": {},
    r"-:.*enorm": {},
    r"-:.*hnorm": {},
    r"-:.*eh_proj": {},
    r"-:.*input_layernorm": {},
    }
```
The layer with index 0 does not have moe modules, so all modules excluded from quantization, so the `subset` len is `0`.
And as variables are not defined the forward_replay fail with:
```
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/models/base.py", line 1016, in quantize
    result = module_looper.loop(
        backend=backend,
        fail_safe=self.quantize_config.fail_safe,
    )
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/looper/module_looper.py", line 1156, in loop
    return self._loop_impl(fail_safe=fail_safe, **kwargs)
[...]

PR #2161, opened 2025-11-01
run forward pass even for empty subset to produce correct layer outputs

avtc:
@Qubitium 
This is a simplified fix for #2116, maybe it is not optimal, but it works.

The suggested solution is too hard for me to make it properly:
> If a full layer has no module to quantize a simple forward() is enough and output is captured to be used as next layer's input. So one pass forward (entire layer simple forward wihout need of dealing with subset loops and micro forward loops, just full layer, usally XXXDecodeLayer.forward(). So output = current_layer.forward() is enough or sometimes just calling the layer callable like layer() which same as layer.forward().

> Assume layer 2 has no modules to quantize. At beginniing loop for layer 2, we have layer_output from completed forward_replay() of layer 1. Then pass this to layer 2 (as a whole) as layer_input and store ouput, then immediately loop to layer 3 without any further subset work that is only necessary if we need to quantize part of a layer.

And it looks it is correct for the case when one or more subsets of a layer are excluded from quantization.

### DIFF
--- a/gptqmodel/looper/stage_subset.py
+++ b/gptqmodel/looper/stage_subset.py
@@ -88,6 +88,29 @@ def run_subset_stage(
         layer_module=module,
     )
 
+    # TODO FIXME: If a full layer has no module to quantize a simple forward() is enough and output is captured 
+    # to be used as next layer's input. So one pass forward (entire layer simple forward wihout need of dealing 
+    # with subset loops and micro forward loops, just full layer, usally XXXDecodeLayer.forward(). 
+    # So output = current_layer.forward() is enough or sometimes just calling the layer callable like layer() 
+    # which same as layer.forward().
+    #
+    # Assume layer 2 has no modules to quantize. At beginniing loop for layer 2, we have layer_output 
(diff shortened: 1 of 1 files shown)


gptqmodel/looper/stage_subset.py
```diff
@@ -88,18 +88,6 @@ def run_subset_stage(
...
-        if logger.isEnabledFor(logging.DEBUG):
-            logger.debug(
-                "StageSubset: layer=%s subset=%s/%s processor=%s produced empty subset (names=%s)",
-                layer_index,
-                subset_index + 1,
-                subset_total,
-                processor_name,
-                subset_names,
-            )
-        return SubsetStageResult(processed_subset={}, layer_inputs=layer_inputs, forward_context=None)
```
Qubitium, line 101: @avtc  Can you un-delete this block but comment out the code instead and add a # TODO FIXME part to show we need to fix this. Thanks.

Reason I ask is this is may not be the final fix, pending future pr, and I don't want to lose the empty state check location in code.

PR #2163, opened 2025-11-01
Retry partial.to to fix accelerate invalid argument error for first moe layer for >4 GPU setups

avtc:
It works for me with 8x3090 from the first attempt.
image

I had the issue on the first layer with experts similar to one we overcome in early version with lock, for me retry works well, as it thrown only on first layer, do not see much reason in a lock there.

The original error trace that is fixed by PR:
```
Traceback (most recent call last):j in layer      [1 of 45] ████░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░| 0:01:02 / 0:23:46 [2/46] 4.3%
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/utils/threadx.py", line 484, in _run
    result = fn(*args, **kwargs)
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/looper/module_looper.py", line 1608, in _process_on_worker
    proc.process(module=nm)
    ~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/looper/gptq_processor.py", line 162, in process
    wq, q_scales, q_zeros, q_g_idx, duration, avg_loss, damp_percent, nsamples = g.quantize()
                                                                                 ~~~~~~~~~~^^
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
    return func(*args, **kwargs)
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/quantization/gptq.py", line 602, in quantize
    self.finalize_hessian(target_device=target_device)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/quantization/gptq.py", line 520, in finalize_hessian
    self._materialize_global_hessian(target_device=target_device)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/venvs/gptqmodelt/lib/python3.13t/site-packages/gptqmodel/quantization/gptq.py", line 503, in _materialize_global_hessian
    tmp = partial.to(device=result_accum.device, dtype=torch.float32)
torch.AcceleratorError: CUDA error: invalid argument
```

### DIFF
--- a/gptqmodel/quantization/gptq.py
+++ b/gptqmodel/quantization/gptq.py
@@ -500,23 +500,19 @@ class GPTQ:
 
             for partial_device, partial in self._device_hessian_partials.items():
                 if partial.device != result_accum.device or partial.dtype != torch.float32:
+                    # TODO FIXME multi-3090 using P2P is revaling an issue where result_accum and/or partial is not ready for consolidation on the main thread
+                    # when parials are calculated on the individual 
                     try:
                         tmp = partial.to(device=result_accum.device, dtype=torch.float32)
                         result_accum.add_(tmp)
                         del tmp
                     except:
                         log.warn(f"Quantization: Module `{self.name}` -> Retry 1/2 partial.to in 0.5s")
-                        time.sleep(0.5)
-                        try:
-                            tmp = partial.to(device=result_accum.device, dtype=torch.float32)
-                            result_accum.add_(tmp)
-                            del tmp
-                        except:
-                            log.warn(f"Quantization: Module `{self.name}` -> Retry 2/2 partial.to in 0.5s")
-                            time.sleep(0.5)
-                            tmp = partial.to(device=result_accum.device, dtype=torch.float32)
-                            result_accum.add_(tmp)
-                            del tmp
+                        time.sleep(0.25)
+                        tmp = partial.to(device=result_accum.device, dtype=torch.float32)
+                        result_accum.add_(tmp)
+                        del tmp
+                       
                 else:
                     result_accum.add_(partial)
 


Qubitium:
@avtc  Amazing simple fix. Looks like the nvidia gpu at a very low level has a internal bg thread that does memory sweep and cleanup! 0.5 seconds delay. Genius! Btw, can you remoe the second retry? I see from the logs that only 1 retry is enough. Is second retry necessary?

Qubitium:
@avtc  Another question, have you tried zero delay os a time delay is absolutely required for this 3090 oom?

gptqmodel/quantization/gptq.py
```diff
@@ -500,9 +500,23 @@ def _materialize_global_hessian(self, target_device: Optional[torch.device] = No
...
+                        try:
+                            tmp = partial.to(device=result_accum.device, dtype=torch.float32)
+                            result_accum.add_(tmp)
+                            del tmp
+                        except:
+                            log.warn(f"Quantization: Module `{self.name}` -> Retry 2/2 partial.to in 0.5s")
+                            time.sleep(0.5)
+                            tmp = partial.to(device=result_accum.device, dtype=torch.float32)
+                            result_accum.add_(tmp)
+                            del tmp
```
Qubitium: @avtc  Try this simpler change and see if it works. This may be another thread/stream sync issue. 

```py
torch.cuda.synchronize(result_accum.device)
tmp = partial.to(device=result_accum.device, dtype=torch.float32)
result_accum.add_(tmp)
```

Qubitium: Also run with env to get more debugging info on crash from cuda:

```
export CUDA_LAUNCH_BLOCKING=1
export TORCH_SHOW_CPP_STACKTRACES=1
```

avtc: @Qubitium 
This particular issue is not connected to OOM, but to initialization of GPUs by Accelerate or something like that. 

Sorry, but it looks like today I cannot reproduce the error so cannot test different solutions properly. (Maybe need to restart server few times...)

`export CUDA_LAUNCH_BLOCKING=1` this leads to deadlock on start (about from the time when data parallel was added +-), that how it looks:
image

Sometimes similar error I have with vllm and for it `export CUDA_LAUNCH_BLOCKING=1` helps.

I have tried `torch.cuda.synchronize(result_accum.device)` and syncronize via accelerate and it did not help but that was long time ago.
I have tried retry immediately without delay and it sometimes works, sometimes not, so to be 100% sure I have added a small delay. The error is thrown only for the first layer with experts, so will not cause slowdown.

Qubitium:
@avtc  Thanks for the investigation and merged. I made a minor adjust to remove the second retry and reduced sleep to 250 ms.

PR #2171, opened 2025-11-03
[FIX] AWQ MoE

ZX-ModelCloud:


Qubitium:
@ZX-ModelCloud  Let's refractor this into model def. We need this fix for all future moes.

PR #2175, opened 2025-11-04
adjust retry partial.to

avtc:
@Qubitium Hi!
I was not able to proceed without second retry, so adjusting back. This is the first MoE layer with experts for GLM-4.5-Air with 1534 samples, "balanced" strategy, on 8 x 3090.

image

### DIFF
--- a/gptqmodel/quantization/gptq.py
+++ b/gptqmodel/quantization/gptq.py
@@ -575,13 +575,13 @@ class GPTQ:
                     try:                    
                         result_accum.add_(partial.to(device=result_accum.device, dtype=torch.float32))
                     except:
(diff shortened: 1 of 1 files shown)


Qubitium:
@avtc  Adjust first delay to 0.25s  (250ms) and second delay to 0.75s. Same total delay but makes first retry  2x as fast which may satisfy 99% of the cases. 

See if you can adjust 0.25 even lower and increase seoncond retry timeout to minimize first retry so it satisify 90% of the cases. Like 0.125 first retry and 0.5s 2nd retry etc. 

Run the values and check maybe 4-5 layers.

avtc:
will check tomorrow, I think it will also work

PR #2182, opened 2025-11-06
Make torch fused op compilable

jiqing-feng:
After this change, the torch fused op can combine with the whole model torch.compile. It could bring 4x speed-up when input_size > 128 on Intel Xeon CPU.


Qubitium:
@jiqing-feng  Wow! Awesome. Thank you!

PR #2190, opened 2025-11-08
AWQ Torch Fused Kernel

Qubitium:
@jiqing-feng  AWQ torch fused kernel based on your awesome work on GPTQ torch fused should be ready soon. Passing CPU tests. Once XPU tests passes, it will be merged.


Qubitium:
xpu tests passing

PR #2249, opened 2025-12-10
Refractor Kernel External Dependency Validation

LRL2-ModelCloud:


gptqmodel/nn_modules/qlinear/marlin.py
```diff
@@ -208,12 +209,15 @@ def __init__(
     #
     #     super().optimize()
 
+
     @classmethod
-    def validate(cls, **args) -> Tuple[bool, Optional[Exception]]:
+    @lru_cache(maxsize=1)
```
Qubitium: @LRL2-ModelCloud  check if the python @lru_cache decorator is using clss name as key. If so, maxsize=1 is wrong and it should be something very high like 100/1024 that holds all kernels.

Issue #2272, opened 2025-12-15
[BUG] Failed to load a gptq_v2 model with bitblas backend


xxxxyu:
**Describe the bug**

When I attempt to load a model in "gptq_v2" format with the "bitblas" backend, it causes a shape mismatch error:

<img width="2284" height="598" alt="Image" src="https://github.com/user-attachments/assets/96902bed-0351-49c4-8b6c-3e1f016c03e1" />

**Software Info**

Name: GPTQModel
Version: 5.4.0

<details>
<summary>See complete info</summary>

```
Name: GPTQModel
Version: 5.4.0
Summary: Production ready LLM model compression/quantization toolkit with hw accelerated inference support for both cpu/gpu via HF, vLLM, and SGLang.
Home-page: https://github.com/ModelCloud/GPTQModel
Author: 
Author-email: ModelCloud <qubitium@modelcloud.ai>
License-Expression: Apache-2.0
Location: /home/lixiangyu/repos/hisense/vllm/gptqv2/lib/python3.12/site-packages
Requires: accelerate, datasets, device-smi, dill, hf_transfer, huggingface_hub, logbar, maturin, numpy, packaging, pillow, protobuf, pyarrow, pypcre, random_word, safetensors, threadpoolctl, tokenicer, torch, torchao, transformers
Required-by: 
---
Name: torch
Version: 2.8.0+cu128
Summary: Tensors and Dynamic neural networks in Python with strong GPU acceleration
Home-page: https://pytorch.org/
Author: PyTorch Team
Author-email: packages@pytorch.org
License: BSD-3-Clause
Location: /home/lixiangyu/repos/hisense/vllm/gptqv2/lib/python3.12/site-packages
Requires: filelock, fsspec, jinja2, networkx, nvidia-cublas-cu12, nvidia-cuda-cupti-cu12, nvidia-cuda-nvrtc-cu12, nvidia-cuda-runtime-cu12, nvidia-cudnn-cu12, nvidia-cufft-cu12, nvidia-cufile-cu12, nvidia-curand-cu12, nvidia-cusolver-cu12, nvidia-cusparse-cu12, nvidia-cusparselt-cu12, nvidia-nccl-cu12, nvidia-nvjitlink-cu12, nvidia-nvtx-cu12, setuptools, sympy, triton, typing-extensions
[...]

PR #2281, opened 2025-12-16
Fix bitblas support for gptq_v2 format

xxxxyu:
Fix https://github.com/ModelCloud/GPTQModel/issues/2272.

I've added a test that requires downloading a 1 GB checkpoint (feel free to remove it).


Qubitium:
@xxxxyu  LGTM. Thanks!

PR #2293, opened 2025-12-19
[Feature] Use FailSafe Naive Quantization when GPTQ fails due to MoE uneven routing

ZX-ModelCloud:


Qubitium:
RTN = round to nearest, naive but low quality quantization that is better than what we do currently if module cannot be quantized due to low calibration data/routing.

Qubitium:
Switch to `midpoint` which is measured to be better under all conditions than `rtn`.

Issue #1682, opened 2025-08-15
[BUG]Error when quantizing Qwen2.5-Omni model using audio data as calibration dataset


Nortinwell:
**Describe the bug**
When attempting to quantize a Qwen2.5-Omni model using audio data as the calibration dataset, the process fails with a StopIteration exception during the dataset preparation phase.

  File "D:\Miniconda3\condabin\qwen25omni_py312\Lib\site-packages\gptqmodel\models\definitions\base_qwen2_5_omni.py", line 111, in prepare_dataset
    inputs = processor(
             ^^^^^^^^^^
  File "D:\Miniconda3\condabin\qwen25omni_py312\Lib\site-packages\transformers\models\qwen2_5_omni\processing_qwen2_5_omni.py", line 198, in __call__
    text = self.replace_multimodal_special_tokens(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Miniconda3\condabin\qwen25omni_py312\Lib\site-packages\transformers\models\qwen2_5_omni\processing_qwen2_5_omni.py", line 241, in replace_multimodal_special_tokens
    sample = sample.replace(self.audio_token, "<|audio_placeholder|>" * next(audio_lengths), 1)
                                                                        ^^^^^^^^^^^^^^^^^^^
StopIteration

**GPU Info**

+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 531.18                 Driver Version: 531.18       CUDA Version: 12.1     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                      TCC/WDDM | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf            Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 3080 T...  WDDM | 00000000:01:00.0  On |                  N/A |
| N/A   51C    P8               21W /  N/A|   1209MiB / 16384MiB |      1%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

**Software Info**

Operation System/Version + Python Version

- OS : Windows
- CUDA : 12.4
- Transformers version : 4.54.0.dev0
- GPTQModel version : Latest

**To Reproduce**

[...]

PR #2309, opened 2025-12-23
Support for Qwen2.5-Omni calibration data includes audio.

ChenShisen:
Environment dependencies: librosa>=0.11.0, audioread>=3.1.0

Bug fix: https://github.com/ModelCloud/GPTQModel/issues/1682

### DIFF
--- a/gptqmodel/models/definitions/base_qwen2_5_omni.py
+++ b/gptqmodel/models/definitions/base_qwen2_5_omni.py
@@ -31,7 +31,7 @@ class BaseQwen2_5_OmniGPTQ(BaseQModel):
 
     pre_lm_head_norm_module = "thinker.model.norm"
 
-    require_pkgs = ["audioread>=3.1.0", "librosa>0.11.0"]
+    require_pkgs = ["audioread>=3.1.0", "librosa>0.11.0", "av>=16.0.1"]
 
     module_tree = [
         "thinker",


requirements.txt
```diff
@@ -19,3 +19,5 @@ pyarrow>=21.0
 dill>=0.3.8
 torchao>=0.14.1
 kernels>=0.11.1
+audioread>=3.1.0
+librosa>0.11.0
```
Qubitium: @ChenShisen Move this to model `cls.require_pkgs`

gptqmodel/utils/audio.py
```diff
@@ -0,0 +1,99 @@
...
+SAMPLE_RATE=16000
+def _check_if_video_has_audio(video_path):
+    container = av.open(video_path)
+    audio_streams = [stream for stream in container.streams if stream.type == "audio"]
+    if not audio_streams:
+        return False
+    return True
+
+
+def process_audio_info(conversations: list[dict] | list[list[dict]], use_audio_in_video: bool):
```
Qubitium: remove the global var `SAMPLE_RATE=16000` and move into process_audio_info as paramter with 16k default. 

```
def process_audio_info(conversations: list[dict] | list[list[dict]], use_audio_in_video: bool, sample_rate: int = 16000):
```

gptqmodel/utils/audio.py
```diff
@@ -0,0 +1,99 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# Contact: qubitium@modelcloud.ai, x.com/qubitium
+
+import base64
+from io import BytesIO
+
+import audioread
+import av
+import librosa
+import numpy as np
```
Qubitium: need to add try/catch here since `audioread`/`librosa` is optional pkg requirement

gptqmodel/utils/audio.py
```diff
@@ -5,23 +5,44 @@
...
+    AV_AVAILABLE = False
+
+try:
+    import librosa
+    LIBROSA_AVAILABLE = True
+except ImportError:
+    LIBROSA_AVAILABLE = False
+
+AUDIOREAD_INSTALL_HINT = "audioread not installed. Please install via `pip install audioread>=3.1.0`."
+AV_INSTALL_HINT = "av not installed. Please install via `pip install av>=16.0.1`."
```
Qubitium, line 29: @ChenShisen Do we need to add `av` to `require_pkgs`?

Qubitium:
@ChenShisen Everything looks good. Please run your model test again to make sure there is no regressions, and let me know when I can merge.

ChenShisen:
Based on your suggestions, the following changes have been made:
    1.Python dependencies are defined within the class (cls).
    2.try/except blocks are used so that checks for audioread, librosa, and av only occur where these dependencies are actually needed.
    3.sample_rate is now a default parameter of the process_audio_info function, with a default value of 16000.

PR #2319, opened 2025-12-29
fix hf kernel

jiqing-feng:
Without static method, this function will treat self as the 1st input args and raise error like:
`*** TypeError: gemm_int4_forward() takes 5 positional arguments but 6 were given`


jiqing-feng:
Hi @Qubitium . Please review this change. Thanks!

PR #2294, opened 2025-12-20
feat: ability to pause/resume quantization via 'p' key

avtc:
Hello, @Qubitium, I have encountered the need to pause long running quantization to be able to turn off a server for night-time. Hope you will like this feature.

Purpose: Ability to pause long-running quantization, so server can be hibernated and then continue its work later on.

Status bar presentation, when running:
image

Pause requested via 'p' or 'Pause/Break' key:
image

Paused at the layer end:
image

PR #2327, opened 2026-01-02
chery pick: attempt to fix terminal state after pause/resume handlers

avtc:
@Qubitium 
https://github.com/ModelCloud/GPTQModel/pull/2294#issuecomment-3688734739

Fix for: looks like this sometimes (after crashes) leads to the state where terminal no longer prints input to the screen (symbols are invisible but after typing and Enter they are accepted)

fix works almost every time :)

### DIFF
--- a/gptqmodel/utils/threadx.py
+++ b/gptqmodel/utils/threadx.py
@@ -18,6 +18,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Un
 
 import torch
 
+from .pause_resume import _restore_terminal_settings_on_exit
 
 try:
     from device_smi import Device  # type: ignore
@@ -441,6 +442,7 @@ class _DeviceWorker:
         except Exception:
             pass
         try:
+            _restore_terminal_settings_on_exit()
             os._exit(1)
         except Exception:
             # Last resort if os._exit is unavailable for some reason.


ZX-ModelCloud:
I tested it, and `_restore_terminal_settings_on_exit()` doesn't seem to be called when an exception is raised; in this case, the terminal still doesn't print input.
However, `_restore_terminal_settings_on_exit()` is called during a `KeyboardInterrupt`.

Qubitium:
@avtc  Was this fix tested in which terminal app on which linux os?  This still looks like some terminal envs/windows have different setup causing the fix to not take affect. @ZX-ModelCloud  Test this in a real terminal and not via ide emulated terminal. Can test under real ssh terminal and tmux terminal inside ssh.

ZX-ModelCloud:
> @avtc Was this fix tested in which terminal app on which linux os? This still looks like some terminal envs/windows have different setup causing the fix to not take affect. @ZX-ModelCloud Test this in a real terminal and not via ide emulated terminal. Can test under real ssh terminal and tmux terminal inside ssh.

I just tested it on the **Konsole** terminal and got the same result.

avtc:
Will check

avtc:
@Qubitium @ZX-ModelCloud please test again. I have tested in Kubuntu terminal and it works now (and restores cursor as well). But do not have ability to test in Windows.

ZX-ModelCloud:
> @Qubitium @ZX-ModelCloud please test again. I have tested in Kubuntu terminal and it works now (and restores cursor as well). But do not have ability to test in Windows.

I retested it, and it's OK.
Additionally, I fixed a bug where the `atexit` callback was not called when a thread raised an exception.

ZX-ModelCloud (approved):

Qubitium (approved):

PR #2332, opened 2026-01-06
Fix NaN error in AWQ quantization for Qwen3

13pathak:
## Description

This PR addresses numerical instability issues encountered when quantizing certain models (specifically Qwen3) with AWQ. The errors manifested as `NaN` values during the `_compute_best_scale` phase, caused by zero-valued channels in weights/activations and unstable scale normalization.

## Changes

* **Clamped Means:** Added `clamp(min=1e-8)` to both `w_mean` and `x_mean` calculations to prevent invalid values in subsequent power operations (e.g., `pow(1 - ratio)`).
* **Stable Scale Normalization:** Replaced the direct `sqrt()` normalization with a conditional check. If the scale product is extremely small ( `< 1e-8` ) or max scale is small ( `< 1e-6` ), it now defaults to a safe fallback instead of risking division by zero or underflow errors.

## Impact

* Resolves `NaN` quantization failures for Qwen3-8B-Base and potentially other models with sparse activations or specific weight distributions.
* Ensures `best_ratio` is correctly found, preventing the `Exception` raised when `best_ratio` remains `-1`.

### DIFF
--- a/gptqmodel/looper/awq_processor.py
+++ b/gptqmodel/looper/awq_processor.py
@@ -725,6 +725,7 @@ class AWQProcessor(LoopProcessor):
             w_mean = torch.zeros(num_channels, dtype=weight_dtype, device=weight_device)
         else:
             w_mean = (w_sum / row_count).to(weight_dtype)
+            w_mean = w_mean.clamp(min=1e-8)
 
         # [STEP 2]: Compute per-channel mean of the input activation with chunking
         # Stream directly on the source device to avoid creating full CPU copies while still enforcing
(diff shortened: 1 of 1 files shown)


Qubitium:
@13pathak This is nothing technically wrong with the code but permise that triggered the condition that require this clamp fix may actually be an internal lifecycle bug in gptqmodel. For example, having any large number of weights as 0 is close to nil in a real model.  We are checking for a core fix. 

Meaning, this fix may fix your errors, but your model is actually not doing the correct awq quantization. There is something wrong with this model happening/triggering at layer 4 (index 3).

PR #2334, opened 2026-01-07
Device check

jiqing-feng:
This PR fix:
1. When device_map is auto, we shoud return the accelerator type instead of None, which will auto select cuda.
2. Fix the linear mode (train or inference) for the very first time and never change it later.

Hi @Qubitium . Would you please review this PR? Thanks!

The reason I want to make this change is because of the peft test failed: [test_adalora_causalLM](https://github.com/huggingface/peft/blob/main/tests/test_gptqmodel.py#L293)


jiqing-feng:
Convert to draft for more checks.

jiqing-feng:
Hi @Qubitium . There are too many tests in the repo. What kind of tests should I run?

Qubitium (approved):

Qubitium:
@jiqing-feng LGTM. Thanks for the fix! We will add a unit test to check to validate the `device_map` for future regression. I don't think we have one to only test `device_map` normalization.

PR #2347, opened 2026-01-09
Re-apply moe routing bypass

ZX-ModelCloud:


Qubitium:
@avtc  We are going to do with this PR.  Full revert actually removed multiple bug fixes/changes so we only want to do a partial revert. 

It's hard to tell from this PR commit tree where the bug is but @ZX-ModelCloud  was able to isolate to this block of removed code here: https://github.com/ModelCloud/GPTQModel/pull/2346/changes/60e2fceb970c44f360aec425b7ca971dded8c3c5. This PR applies the moe bypass without that block of code. 

Basically overridng `NamedModule.forward` was the bug and we may not even need those code. It appears to be useless and will actually break forwarding for awq. Currently ci testing this PR to check for both gptq/awq tests.

PR #2349, opened 2026-01-10
feat: moe-router-bypass-batch-size

avtc:
@Qubitium Hi, this feature adds ability to process moe weights in chunks in MoeRoutingBypass mode. Allowing to quantize large MoE models with less VRAM and devices.

Currently for each expert weight in GPTQ the hessian accumulator created during forward pass, which for example for GLM-4.5-Air is around 17Gb for up/gate moe modules for one layer. Processing expert weights in chunks require less hessian accumulator matrices and thus VRAM.

Example option usage:
```
moe={'routing': {'class': 'ExpertsRoutingBypass', 'batch_size': 50}}
```


Qubitium:
@avtc Do you have some rough numbers for vram saving in your setup with and without the PR?

avtc:
@Qubitium Hi,
Here are screens from quantizing a GLM-4.5-Air - `layer 0` - non moe, `layer 1+` - moe:
Calibration data: 2320 samples - 769K tokens.
5x3090 with first one excluded from compute: `compute_device_filter=lambda devices: [d for d in devices if d.index != 0]`
`VramStrategy=Exclusive` (i.e. data parallel).

Without batching experts, forward of moe up/gate modules on the `layer 1`:
image
Results in CUDA OOM in a while after screenshot.

With `batch_size = 128`, same stage:
image
It works, we batch 128 modules from 256 (128 x up/gate) in one forward/quantize.

With `batch_size = 32`, same stage for comparison:
image

Qubitium:
> @Qubitium Hi, Here are screens from quantizing a GLM-4.5-Air - `layer 0` - non moe, `layer 1+` - moe: Calibration data: 2320 samples - 769K tokens. 5x3090 with first one excluded from compute: `compute_device_filter=lambda devices: [d for d in devices if d.index != 0]` `VramStrategy=Exclusive` (i.e. data parallel).
> 
> Without batching experts, forward of moe up/gate modules on the `layer 1`: image Results in CUDA OOM in a while after screenshot.
> 
> With `batch_size = 128`, same stage: image It works, we batch 128 modules from 256 (128 x up/gate) in one forward/quantize.
> 
> With `batch_size = 32`, same stage for comparison: image

Looks good!  We will work on this PR after a newly `awq` discovered regression is fixed.

tests/module_tree/test_subset.py
```diff
@@ -300,6 +300,19 @@ def test_stage_subset_early_stop_and_callbacks():
...
+        full=full_modules,
+        is_lm_head_module=False,
+        layer_index=0,
+        layers_prefix="layers",
+        names=subset_names,
+        processor=processor,
+        failsafe=False,
+        layer_module=mini_layer,
+    )
+
```
Qubitium: The second subset overrides the first.

tests/test_moe_config.py
```diff
@@ -63,6 +63,19 @@ def assert_results(model, target_class, moe_config: MoEConfig):
...
+        moe = qcfg.meta_get("moe")
+        assert moe is not None, "MoE metadata should be present"
+        assert moe["routing"]["class"] == moe_dict["routing"]["class"]
+        # Verify routing-specific fields
+        if moe_dict["routing"]["class"] == "ExpertsRoutingBypass":
+            if "batch_size" in moe_dict["routing"]:
+                assert moe["routing"]["batch_size"] == moe_dict["routing"]["batch_size"]
+        elif moe_dict["routing"]["class"] == "ExpertsRoutingOverride":
+            assert moe["routing"]["num_experts_per_tok"] == moe_dict["routing"]["num_experts_per_tok"]
+
```
Qubitium: @ZX-ModelCloud  Let's remove the json parsing/decoding of a python class. I would to avoid this for now.

gptqmodel/quantization/config.py
```diff
@@ -557,6 +564,64 @@ def _parse_smooth_method(setting: Any) -> Optional[SmoothMethod]:
...
+def _build_moe_routing_from_dict(payload: Dict[str, Any]) -> BaseMoERouting:
+    """
+    Build MoE routing object from dictionary.
+
+    Expected format:
+    {
+        "class": "ExpertsRoutingBypass" | "ExpertsRoutingOverride",
+        "batch_size": int (optional, for ExpertsRoutingBypass),
+        "num_experts_per_tok": int | str (for ExpertsRoutingOverride)
+    }
```
Qubitium: @ZX-ModelCloud  per previous comment. Let's remove this json python class decode for now.

Qubitium (changes requested):
Cleanup

Qubitium:
@avtc  5.7.0 took much longer to push out due to small bugs we keep finding/fixing in CI regression tests. Should be out soon so this can be merged.

PR #2365, opened 2026-01-19
Fix evalplus output filename mismatch

juraev:
## Summary
- Fix compatibility with evalplus 0.4.0+ which changed output filename format from `_eval_results.json` to `.eval_results.json`
- Match evalplus's behavior by checking both filename patterns with fallback

## Test plan
- [x] Verify syntax is valid
- [ ] Test with existing legacy `_eval_results.json` files (backwards compatible)
- [ ] Test with new `.eval_results.json` files from evalplus 0.4.0+

### DIFF
--- a/gptqmodel/utils/eval.py
+++ b/gptqmodel/utils/eval.py
@@ -148,8 +148,11 @@ def evalplus(
              greedy=True)
 
     if output_file is None:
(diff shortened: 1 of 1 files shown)


Qubitium (approved):

Qubitium:
@juraev Thank you for the compat fix!

PR #2386, opened 2026-01-27
support video input for quantization

techshoww:
I have migrated the process_vision_info function from the qwen_vl_utils library to here, to support video data input during quantization.


Qubitium:
@techshoww Thank you for the contribution! Everything looks good. It will be merged once we release 5.7.0 wheels which is under final testing. It will be merged into `main` post 5.7.0 cut.

PR #2403, opened 2026-02-05
Fix tokenizer_class incompatibility with transformers 5.0

juraev:
## Summary
- On transformers 5.0, `save_pretrained()` writes `"TokenizersBackend"` as the `tokenizer_class` in `tokenizer_config.json`
- The existing fixup code (for [huggingface/transformers#35832](https://github.com/huggingface/transformers/issues/35832)) appends `"Fast"` to any non-Fast class name, producing `"TokenizersBackendFast"` — a class that doesn't exist in any transformers version
- This breaks all downstream consumers (vLLM, evalplus, etc.) with: `ValueError: Tokenizer class TokenizersBackendFast does not exist or is not currently imported.`
- Fix: read `tokenizer_class` from the **source model** instead of from the `save_pretrained` output, preserving a backwards-compatible model-specific class name (e.g. `LlamaTokenizerFast`) regardless of which transformers version does the saving

## Test plan
- [ ] Quantize a model (e.g. `deepseek-ai/deepseek-coder-1.3b-instruct`) using transformers 5.0 and verify `tokenizer_config.json` has a model-specific class name (e.g. `LlamaTokenizerFast`), not `TokenizersBackendFast`
- [ ] Verify the quantized model loads correctly on both transformers 4.x and 5.x
- [ ] Verify downstream tools (vLLM, evalplus) can load the quantized model without errors

### DIFF
--- a/gptqmodel/models/writer.py
+++ b/gptqmodel/models/writer.py
@@ -498,16 +498,22 @@ def ModelWriter(cls):
         if self.tokenizer:
             self.tokenizer.save_pretrained(save_dir)
 
-            # fixed this issue: https://github.com/huggingface/transformers/issues/35832
-            saved_tokenizer_config = get_tokenizer_config(save_dir)
-            config_tokenizer_class = saved_tokenizer_config.get("tokenizer_class")
-            # if the tokenizer is fast, but the tokenizer_config.json does not have Fast suffix, add "Fast" suffix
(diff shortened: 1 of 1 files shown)


Qubitium:
@juraev  Thank you for the HF Transformers 5.0 compat bug fix!

PR #2405, opened 2026-02-07
Add EXAONE4 support 

namgyu-youn:
As mentioned in title, add support for EXAONE4 series. Source: https://huggingface.co/LGAI-EXAONE/EXAONE-4.0-1.2B


Qubitium:
@namgyu-youn Thank you for the Exaone4 support! GPTQModel was the first quantizer to support the first ExaOne model and we are glad we have contributors continuning this support trend. This code have very low regression risk so we will merge into the 5.7.0 tree asap.

PR #2411, opened 2026-02-10
[FIX] module_tree tests

ZX-ModelCloud:


Qubitium:
This was reapplying a fix that was reverted due to regression but reverting the revert missed the original patch fix.

PR #2422, opened 2026-02-20
Feature/LLada2 support: Block Diffusion LLM

blazingbhavneek:
Adding support of Diffusion LLM: https://huggingface.co/inclusionAI/LLaDA2.1-mini

Short summary: Architecture is same as normal LLM except 4D attention mask which was monkey patched after loading

Tests: I need some time to replicate some benchmarks on my PC (GPU Poor), but I was able to test on my work GPUs (Nvidia H200s) and accuracy loss was 2-3% (Calibrated with 1000 samples wiki dataset, 4 bit quant)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -865,7 +865,7 @@ class CachedWheelsCommand(_bdist_wheel):
         if FORCE_BUILD or xpu_avail:
             return super().run()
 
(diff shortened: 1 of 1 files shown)


Qubitium:
@blazingbhavneek LGTM! Thank you for the contribution! I will add a ci test to make sure it survive future code changes.

PR #2452, opened 2026-03-09
rm sym in dequantize_gemm

jiqing-feng:
rm sym in dequantize_gemm


jiqing-feng:
Hi @Qubitium . Would you please review this PR? I forgot to remove the unexpected parameter of `dequantize_gemm` function

Issue #2440, opened 2026-03-04
[Transformers 5.2 Compat] AWQ quantization fails for Qwen3.5 with device mismatch in rotary embeddings


erm14254:
AWQ quantization crashes on Qwen3.5-27B at layer 3 (self_attn layer) with a device mismatch error. The rotary embeddings tensor is on CPU while other tensors are on CUDA. GPTQ quantization works fine for the same model.

NVIDIA-SMI 595.71
Driver Version: 595.71
CUDA Version: 13.2 
Windows 11 Pro
Python 3.12.10
torch: 2.10.0+cu130
transformers: 5.2.0
accelerate 1.12.0
GPTQModel 5.7.1 (commit 83a9a06ee0d9e97f84f68f5996589a642b1e5080)

**To Reproduce**

from gptqmodel import GPTQModel
from gptqmodel.quantization import QuantizeConfig, FORMAT, METHOD
from datasets import load_dataset

quantize_config = QuantizeConfig(
    bits=4,
    group_size=128,
    format=FORMAT.LLM_AWQ,
    quant_method=METHOD.AWQ,
)

dataset = load_dataset("allenai/c4", "en", split="train", streaming=True)
calibration_data = []
for sample in dataset:
    if len(sample["text"]) > 2000:
        calibration_data.append(sample["text"])
    if len(calibration_data) >= 256:
        break

model = GPTQModel.load(
    "Qwen/Qwen3.5-27B",  # or any Qwen3.5 model
    quantize_config=quantize_config,
    trust_remote_code=True,
)

model.quantize(
    calibration=calibration_data,
    batch_size=1,
)

**Expected behavior**

AWQ quantization should complete successfully, as GPTQ quantization does for the same model.

**Model/Datasets**

- Model: `Qwen/Qwen3.5-27B` (or any Qwen3.5 variant)
- Dataset: `allenai/c4`

**Error Traceback**

RuntimeError: Expected all tensors to be on the same device, but got mat2 is on cuda:0, different from other tensors on cpu (when checking argument in method wrapper_CUDA_bmm)

File "gptqmodel/looper/awq_processor.py", line 1096, in _module_forward
    module_kwargs["position_embeddings"] = rotary(x_for_rotary, pos_for_rotary)

File "transformers/models/qwen3_5/modeling_qwen3_5.py", line 238, in forward
    freqs = (inv_freq_expanded.float() @ position_ids_expanded.float()).transpose(2, 3)

**Additional context**

[...]

PR #2453, opened 2026-03-09
fix awq rotary device mismatch. store per-device copy of rotary

Qubitium:
Fix: https://github.com/ModelCloud/GPTQModel/issues/2440

@erm14254 Please test this PR. If it fixes your issue for good, I will merge.

### DIFF
--- a/gptqmodel/looper/awq_processor.py
+++ b/gptqmodel/looper/awq_processor.py
@@ -386,8 +386,15 @@ class AWQProcessor(LoopProcessor):
                 try:
                     pe = rotary(x_for_rotary, pos_for_rotary)
                     refreshed["position_embeddings"] = pe
-                except Exception:
-                    pass
+                except Exception as exc:
+                    # Cache refresh can proceed without precomputed rotary embeddings; log for debugging.
+                    log.debug(
+                        "AWQProcessor: failed to refresh rotary position_embeddings "
+                        "(seq_len=%s, device=%s): %s",
+                        seq_len,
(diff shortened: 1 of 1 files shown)


Qubitium:
Notes: 

  1. Qwen3.5 in transformers 5.2 changed the attention contract.
     In Qwen3_5TextModel.forward:1313, the model now computes shared position_embeddings = self.rotary_emb(...) once at the model level and passes them into each decoder layer. In Qwen3_5Attention.forward:740, position_embeddings is a direct required input. Older/other models often computed RoPE
     inside the layer, or AWQ did not need to emulate this exact model-level precompute.
  2. Qwen3.5 uses one shared rotary module for the whole model, not one per layer.
     You can see that in Qwen3_5TextModel.init:1301: self.rotary_emb = Qwen3_5TextRotaryEmbedding(...). The decoder layers in Qwen3_5DecoderLayer.init:828 do not own their own rotary module. During quantization, GPTQModel keeps most of the shell model on CPU/meta and only materializes the current
     layer/submodule on GPU, so the shared top-level rotary could stay on CPU while the active attention module was on cuda:0.
  3. AWQ had to emulate full-model behavior while only forwarding one module at a time.
     That logic is in AWQProcessor._module_forward:1091 and AWQProcessor._refresh_forward_kwargs_from_cache:345. Before the fix, AWQ built position_embeddings from the shared root rotary without first ensuring that rotary lived on the same device as the layer input. That is why the crash appeared only
     in AWQ, not in normal full-model forward.

PR #2472, opened 2026-03-11
read dependencies from requirements.txt

CSY-ModelCloud:
@Qubitium


CSY-ModelCloud:
[tool.uv]
torch-backend = "auto"

make sure uv pip install torch will install correct torch version which corespond to local cuda version

PR #2498, opened 2026-03-13
Update HFKernelLinear selection

jiqing-feng:
Update HFKernelLinear selection: restrict optimized int4 GEMM path to CPU-only and require release torch

### Changes

- **Restrict int4 GEMM inference fast-path to CPU**: Added `x.device.type == "cpu"` guard in both `HFKernelLinear.forward()` and `HFKernelAwqLinear.forward()` so the optimized `gemm_int4_forward_kernel` transform is only triggered on CPU tensors. On other devices (e.g. CUDA), the code now falls through to the standard dequantize path.
- **Reject non-release torch builds**: Added `_is_torch_release()` class method and an early check in `validate_once()` that returns `False` with a clear warning when running on nightly / dev / RC builds of PyTorch, since the HF kernels extension requires a stable release.

### DIFF
--- a/gptqmodel/nn_modules/qlinear/gemm_hf_kernel.py
+++ b/gptqmodel/nn_modules/qlinear/gemm_hf_kernel.py
@@ -132,7 +132,21 @@ class HFKernelLinear(PackableQuantLinear):
         self.dequant_dtype = torch.int8
 
     @classmethod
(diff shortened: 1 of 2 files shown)


jiqing-feng:
Hi @Qubitium . Please review this PR. Thanks!

PR #2531, opened 2026-03-16
TorchFused: Add fp32 dtype support & fix AWQ dequantize

jiqing-feng:
- **Add `torch.float32` to `SUPPORTS_DTYPES`** for both `TorchFusedQuantLinear` and `TorchFusedAwqQuantLinear`.
  - Since the underlying int4 fused kernels (`_weight_int4pack_mm_for_cpu` / `_weight_int4pack_mm_with_scales_and_zeros`) only support fp16/bf16, fp32 inputs are automatically cast to bf16 before computation and cast back to fp32 on output.
- **Remove unused `sym` argument** from `awq_weight_dequantize()` call in `TorchFusedAwqQuantLinear`, fixing a call mismatch with `dequantize_gemm()`.

Hi @Qubitium . Please review this PR. Thanks!

### DIFF
--- a/gptqmodel/nn_modules/qlinear/torch_fused_awq.py
+++ b/gptqmodel/nn_modules/qlinear/torch_fused_awq.py
@@ -233,7 +233,6 @@ class TorchFusedAwqQuantLinear(TorchFusedQuantLinear):
             scales=self.scales,
             bits=self.bits,
             group_size=self.group_size,
-            sym=self.sym,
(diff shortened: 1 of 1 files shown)


Qubitium:
@jiqing-feng LGTM! Thank you! btw, main is a little unstable right now as we are fixing/injecting a lot of hf transformer 5.3 compat fixes. Our ref ci will run with transformers 5.3.0 plus torch 2.10.0

PR #2592, opened 2026-03-24
speedup paroquant test

ZX-ModelCloud:


tests/models/model_test.py
```diff
@@ -1117,18 +1121,32 @@ def check_kernel(self, model, expected_kernels):
...
+                    adapter=self.EORA,
+                    vram_strategy=self.VRAM_STRATEGY,
+                    dynamic=self.DYNAMIC,
+                    moe=self.MOE_CONFIG,
+                    offload_to_disk=self.OFFLOAD_TO_DISK,
+                )
+            elif self.METHOD == METHOD.PAROQUANT:
+                return ParoQuantQuantizeConfig(
+                    bits=self.BITS,
+                    format=self.BNB_FORMAT,
```
Qubitium: why is paroquant using `self.BNB_FORMAT`?

PR #2421, opened 2026-02-19
Feature/calibration data device

avtc:
@Qubitium Hi, this feature allows specify in config the device where calibration data inputs/outputs will be stored, allowing to use more calibration data samples for quantization, because calibration data can be placed on device different to `cuda:0` which already stores all layer modules.

Before the feature initial calibration data was stored on `CPU` and after first pass it was stored on `DEVICE_0` (cuda:0 usually).
After the feature if `calibration_data_device` is not set initial behavior preserved.
`calibration_data_device` can be set to "cpu", "cuda:1" (or any other torch device), and to "balanced" - in "balanced" mode calibration data distributed between compute devices available: `DEVICE_0` .. `DEVICE_N`

P.S. I have used this feature previously several times but on another old branch. This PR is based on latest master.
Also I have fixed examples in config file for using `moe` parameter, and fixed `sys.abiflags` typo which failed build.

Note: the handling of layer with all modules excluded from quantization was also fixed, as current main code did not do forward replay it seems.

I have run several small tests (few first layers) ensuring nothing fail with `auto_forward_data_parallel` enabled and disabled, on qwen3-30b-a3b with `calibration_data_device` set to `cpu`, `cuda:1`, `balanced` and removed from config.

PR #2608, opened 2026-03-25
Calibration data device

avtc:
@Qubitium Hi, this feature allows specify in config the device where calibration data inputs/outputs will be stored, allowing to use more calibration data samples for quantization, because calibration data can be placed on device different to `cuda:0` which already stores all layer modules.

Before the feature initial calibration data was stored on `CPU` and after first pass it was stored on `DEVICE_0` (cuda:0 usually).
After the feature if `calibration_data_device` is not set initial behavior preserved.
`calibration_data_device` can be set to "cpu", "cuda:1" (or any other torch device), and to "balanced" - in "balanced" mode calibration data distributed between compute devices available: `DEVICE_0` .. `DEVICE_N`

P.S. I have used this feature previously several times but on another old branch. This PR is based on latest master.
Also I have fixed examples in config file for using `moe` parameter, and fixed `sys.abiflags` typo which failed build.

Note: the handling of layer with all modules excluded from quantization was also fixed, as current main code did not do forward replay it seems.

I have run several small tests (few first layers) ensuring nothing fail with `auto_forward_data_parallel` enabled and disabled, on qwen3-30b-a3b with `calibration_data_device` set to `cpu`, `cuda:1`, `balanced` and removed from config.

---
P.S.S: recreated PR from latest main
Original PR: #2421

### DIFF
--- a/gptqmodel/looper/forward_executor.py
+++ b/gptqmodel/looper/forward_executor.py
@@ -278,8 +278,10 @@ class ForwardExecutor:
                 if need_outputs and module_output is not None:
                     primary = module_output[0] if isinstance(module_output, tuple) else module_output
                     # Move output back to the same device where input was stored
-                    # This preserves calibration data placement
-                    primary = move_to(primary, device=input_device)
+                    # This preserves calibration data placement when calibration_data_device is set
+                    calib_device_cfg = self.looper.gptq_model.quantize_config.calibration_data_device
+                    target_device = input_device if calib_device_cfg is not None else cur_layer_device
+                    primary = move_to(primary, device=target_device)
                     outputs.append([primary])
 
                 if module_output is not None:
@@ -534,9 +536,11 @@ class ForwardExecutor:
                 raise RuntimeError("Forward batch returned no output; data-parallel execution produced empty result.")
             primary = module_output[0] if isinstance(module_output, tuple) else module_output
             # Move output back to the same device where input was stored
-            # This preserves calibration data placement
+            # This preserves calibration data placement when calibration_data_device is set
+            calib_device_cfg = self.looper.gptq_model.quantize_config.calibration_data_device
             input_device = layer_inputs[idx][0].device if layer_inputs[idx] else cur_layer_device
-            primary = move_to(primary, device=input_device)
+            target_device = input_device if calib_device_cfg is not None else cur_layer_device
+            primary = move_to(primary, device=target_device)
             ordered_outputs.append([primary])
 
         return ordered_outputs


avtc:
@Qubitium for me integration tests (using real model) fail during loading quantized model step. I have tried to figure out, but with no luck.

The error log:
```
INFO  Pre-Quantized model size: 2098.20MB, 2.05GB                                                                                                                                          
INFO  Quantized model size: 731.80MB, 0.71GB                                                                                                                                               
INFO  Size difference: 1366.40MB, 1.33GB - 65.12%                                                                                                                                          
INFO  | Submodule finalize | 308   | 0.357  | 0.688 | 211.927 | 42.4%  | model.layers.21.mlp.down_proj                       |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Finalize pack      | 154   | 0.327  | 0.745 | 114.806 | 23.0%  | model.layers.21.mlp.gate_proj [module.pack_block]   |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Process quant      | 308   | 1.130  | 0.293 | 90.108  | 18.0%  | model.layers.21.mlp.down_proj                       |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Finalize create    | 154   | 0.004  | 0.443 | 68.165  | 13.6%  | model.layers.21.self_attn.k_proj                    |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Pre-quant forward  | 176   | 0.085  | 0.041 | 7.185   | 1.4%   | model.layers.21:subset4/4                           |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Finalize offload   | 154   | 0.017  | 0.027 | 4.138   | 0.8%   | model.layers.21.mlp.down_proj                       |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Forward hook       | 616   | 0.011  | 0.002 | 1.322   | 0.3%   | model.layers.21.mlp.down_proj                       |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Post-quant replay  | 21    | 0.033  | 0.055 | 1.149   | 0.2%   | model.layers.20:subset4/4                           |                                                             
INFO  +--------------------+-------+--------+-------+---------+--------+-----------------------------------------------------+                                                             
INFO  | Model save         | 1     | 0.778  | 0.778 | 0.778   | 0.2%   | /home/ubuntu/tmp_for_install/pytest-of-ubuntu/pytest-3/test_calibration_data_device_c0/quantized |                
INFO  +--------------------+-------+--------+-------+---------+--------+--------------------------------------------------------------------------------------------------+

INFO  | Turtle reload      | 3     | 0.118  | 0.096 | 0.289   | 0.1%   | auto:LlamaDecoderLayer                                                                           |                
INFO  +--------------------+-------+--------+-------+---------+--------+--------------------------------------------------------------------------------------------------+                
INFO  | Process finalize   | 2     | 0.006  | 0.064 | 0.127   | 0.0%   | tp-pre-pad                                                                                       |                
INFO  +--------------------+-------+--------+-------+---------+--------+--------------------------------------------------------------------------------------------------+                
INFO  | Capture inputs     | 1     | 0.065  | 0.065 | 0.065   | 0.0%   | cache_inputs:LlamaDecoderLayer                                                                   |                
INFO  +--------------------+-------+--------+-------+---------+--------+--------------------------------------------------------------------------------------------------+                
from_quantized: adapter: None
INFO  Loader: Auto dtype (native bfloat16): `torch.bfloat16`                                                                                                                               
INFO  Estimated Quantization BPW (bits per weight): 4.2875 bpw, based on [bits: 4, group_size: 128]                                                                                        
Fetching 0 files: 0it [00:00, ?it/s].00B [00:00, ?B/s]
Fetching 18 files: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 18/18 [00:00<00:00, 85211.59it/s]
Downloading (incomplete total...): 0.00B [00:00, ?B/s]Fatal Python error: Segmentation fault
Download complete: : 0.00B [00:00, ?B/s]                                                                                                                             | 0/18 [00:00<?, ?it/s]
Thread 0x00007b5d3b7fe6c0 (most recent call first):
  File "/home/ubuntu/.pyenv/versions/3.13.7t/lib/python3.13t/threading.py", line 359 in wait
  File "/home/ubuntu/.pyenv/versions/3.13.7t/lib/python3.13t/threading.py", line 659 in wait
  File "/home/ubuntu/git/avtc/GPTQModel/gptqmodel/utils/threadx.py", line 1794 in _janitor_loop
  File "/home/ubuntu/.pyenv/versions/3.13.7t/lib/python3.13t/threading.py", line 994 in run
  File "/home/ubuntu/.pyenv/versions/3.13.7t/lib/python3.13t/threading.py", line 1043 in _bootstrap_inner
  File "/home/ubuntu/.pyenv/versions/3.13.7t/lib/python3.13t/threading.py", line 1014 in _bootstrap
```

Qubitium:
@avtc there is a bug in on the upstream pkgs. known issue with fix pending. do this

pip unistall kernels

u dont need this pkg, for cpu kernels 

if that does not fix it, there maybe other multi threading bugs in play

avtc:
@Qubitium thanks, that helped:
> pip unistall kernels

Edits are allowed for maintainers for this branch. The PR is ready.

gptqmodel/looper/stage_subset.py
```diff
@@ -360,10 +360,9 @@ def build_layer_subset_plans(
             fallback,
             layer_module=module,
         )
-        # Skip empty subsets caused by per-layer structure differences or dynamic
-        # exclusions so execution only sees real work.
-        if subset:
-            subsets.append(subset)
+        # Note: Empty subsets are handled by _handle_empty_subset in run_subset_stage
+        # which skips quantization but still produces consistent events for monitoring
+        subsets.append(subset)
```
Qubitium: @avtc This and subsequent `empty` subset change I don't understand. Why does calibration device placement optimization require this lifecycle change. This is a pretty significant `lifecycel` change as it is going from doing nothing to now doing something special for nothing, which may break awq. 

This single line of change is causing further code to be added to handle this special condition. 

can we remove this empty subset handling all toegether if it was only for logging/debugging purposes?

If it has nothing to do with tensor device placement, let's clean it up.

gptqmodel/looper/stage_inputs_capture.py
```diff
@@ -78,15 +78,39 @@ def cache_inputs(
...
+                balanced_devices = compute_device_filter(all_devices)
+                if not balanced_devices:
+                    balanced_devices = all_devices
+            else:
+                balanced_devices = all_devices
+            data_device = balanced_devices[0] if balanced_devices else cur_layer_device
+        elif calib_device_cfg is not None:
+            data_device = calib_device_cfg
+        else:
+            data_device = cur_layer_device
```
Qubitium, line 110: @avtc At line 86, `cur_layer_device = CPU` is forced `CPU` and line 97 all devices is using `CPU`. 

So my question is, can `data_device` be anything but `CPU` after these hightlighted block part of the code ends?

gptqmodel/looper/stage_subset.py
Qubitium: is it to fix this?

> Note: the handling of layer with all modules excluded from quantization was also fixed, as current main code did not do forward replay it seems.

avtc: Yes

gptqmodel/looper/stage_inputs_capture.py
avtc: I will check tomorrow, looks suspicious

gptqmodel/looper/stage_subset.py
Qubitium: > Yes

Checking this issue right now if it was a regression in the v6.0 refracotr. If subset plan is empty, no more subsets, it should just use the first forward output as input for the next layer if that output cache exists or run a forward pass.

Qubitium: Peding fix in progress. Once validated, it will make this PR much cleaner. 

https://github.com/ModelCloud/GPTQModel/pull/2611

Qubitium: @avtc  fixed this condition in `main` and added very strict full model tests to guard against this condition for future regression. Please merge with main and remove the empty subset processing in the existing code as it should no longer be necessary, this pr should be much cleaner after merge.

note: `fwd_after_process` is renamed to `fwd_replay_after_process`. 

THe merge pr looks like lots of changes but the actual bug fix is in:

https://github.com/ModelCloud/GPTQModel/pull/2611/changes#diff-2ae6705966ac1a84261f843496e8ecd2e1c046052d79e00e478bb6f635b2c0f5

gptqmodel/looper/stage_inputs_capture.py
avtc: @Qubitium 
fixed, and reverted changes to empty_subset

gptqmodel/looper/forward_executor.py
```diff
@@ -274,7 +277,9 @@ def run_single(
 
                 if need_outputs and module_output is not None:
                     primary = module_output[0] if isinstance(module_output, tuple) else module_output
-                    primary = move_to(primary, device=cur_layer_device)
+                    # Move output back to the same device where input was stored
+                    # This preserves calibration data placement
+                    primary = move_to(primary, device=input_device)
```
Qubitium: @avtc I don't see a `calibration_data_device` enabled gate/guard for this code. Will this cause non-enabled to also hit this? Is the intended effect aka optimization unrelated `calibration_data_device`?

gptqmodel/looper/forward_executor.py
```diff
@@ -508,7 +533,10 @@ def _replica_progress(idx: int, total: int, device: torch.device, step: str) ->
             if module_output is None:
                 raise RuntimeError("Forward batch returned no output; data-parallel execution produced empty result.")
             primary = module_output[0] if isinstance(module_output, tuple) else module_output
-            primary = move_to(primary, device=cur_layer_device)
+            # Move output back to the same device where input was stored
+            # This preserves calibration data placement
+            input_device = layer_inputs[idx][0].device if layer_inputs[idx] else cur_layer_device
+            primary = move_to(primary, device=input_device)
```
Qubitium: @avtc The same question as the last block of same logic. If I don't set `calibration_data_device`, this new state change still hits?

avtc: @Qubitium 
Yes, both will execute no matter of calibration data device enabled or not.

The logic behind that to have predictible vram usage for calibration data placement. As right now it initially placed on CPU, then it moved by demand for each weight to the weight device during weight forward pass.

And in _forward_single, _forward_parallel it moved another one time, but not to the weight device, but to layer device (which looks redundant, as weights can be distributed between devices - see balanced mode (vram strategy) how it moves inputs in _forward_single.

Preserving original behavior - place on CPU, and then on first layer move to cuda:0 can be OK for small layers, that fit in vram with all calibration data and H matrices at the same time.

But keeping calibration data on initial device (CPU) will make vram pressure much less in case of large dataset.

Qubitium: > @Qubitium Yes, both will execute no matter of calibration data device enabled or not.
> 
> The logic behind that to have predictible vram usage for calibration data placement. As right now it initially placed on CPU, then it moved by demand for each weight to the weight device during weight forward pass.
> 
> And in _forward_single, _forward_parallel it moved another one time, but not to the weight device, but to layer device (which looks redundant, as weights can be distributed between devices - see balanced mode (vram strategy) how it moves inputs in _forward_single.
> 
> Preserving original behavior - place on CPU, and then on first layer move to cuda:0 can be OK for small layers, that fit in vram with all calibration data and H matrices at the same time.
> 
> But keeping calibration data on initial device (CPU) will make vram pressure much less in case of large dataset.

Got it. I still would like it gated by another control knob:

`return_outputs_to_cache_device` default to `false`

Can't think of a better name for now. We can always rename it later.

Because we need to preserve the original method for compat sakes. On some devices ping-pong tensors back and forth is more costly than doing inference. Power users can enable this coupled with `calibration_data_device` for max vram control.

avtc: Does it make sense to introduce another flag, while calibration data already cover inputs and outputs?

avtc: Ok, i will use calibration_data_device None for backward comp.

Qubitium: > Does it make sense to introduce another flag, while calibration data already cover inputs and outputs?

Yes.I think of it as `calibration_data_device` just optimizes the `cache` device. But these two places live outside of that so you want to save ram by always returning output to cache device, wether it is optimized or not. 

Meaning, if I disable calibration_data_device, I do not want to return to `sender` aka `cache` aka `origin` device by default (original lifecycle).

Qubitium: btw. I am about head to sleep soon so it may be a while before I can reply.

Qubitium: > Meaning, if I disable calibration_data_device, I do not want to return to `sender` aka `cache` aka `origin` device by default (original lifecycle).

Clarification, typo. What I meant was even if I enable or disable `calibration_data_device`, I still want to control if want to enable this "return-to-sender-device" optimization which can help some envs but hurt in others.

Qubitium: Also let's think abut the names which maybe confusing both of us. `calibration_`  before the first layer is really the `calibration` dataset. After that, after the it enters the first layer module, it no longer becomes `calibration` in name (which confuses with the calibration data) but the `activate output` to next module `input`. This activated output is used for calibration data purposes so the namae is not wrong but ithe current name has bad conflict with `calibration` dataset which confuses people even more. lol..

avtc: First layer receives calibration dataset, it then passed to next layers as output from prev layers, but this is still calibration data, i think name is fine. We do not need to separate it to calibration dataset and activation outputs as this will not bring any improvement not to the logic not to the understanding.

Qubitium: `pin_output_to_origin_device` sounds better than my last suggestion for that second control.
`input_device` to replace `calibration_data_device`? what do you think?

There are all good/bad in some ways. I will defer to you.

avtc: I think adding new concept of "return-to-sender-device" will not improve understanding or logic either. If user want to optimize movement calibration data between devices - he will set `calibration_data_device="cuda:0"`, and will use `cuda:0` for forward passes via `auto_forward_data_parallel=False,` - this way calibration data will not be moved between devices at all, but require more VRAM, for small layers and small datasets will be totally fine.

Qubitium: > I think adding new concept of "return-to-sender-device" will not improve understanding or logic either. If user want to optimize movement calibration data between devices - he will set `calibration_data_device="cuda:0"`, and will use `cuda:0` for forward passes via `auto_forward_data_parallel=False,` - this way calibration data will not be moved between devices at all, but require more VRAM, for small layers and small datasets will be totally fine.

I think the only block btween you and I right now is, when `calibration_data_device = None`, I want to avoid any lifecycle deviation, including the lifecycle of the tensors but that two spots did change the logic? So that's why I said to gate that two spots with another control so user has all the options and default lifecycle is also maintained.

avtc: @Qubitium 
added backward compat

Qubitium:
@avtc monitor this `ci` action:

https://github.com/ModelCloud/GPTQModel/actions/runs/23610041993

ONce it pass a few model tests, I will merge. Ignore any errors unrelated to your code but if you see something related, fix.

Qubitium:
CI actually, broken. going to merge.

Qubitium:
@avtc wth. I can't merge. github is freaking. maybe it queued. If it doens't merge, I will merge when I wake up.


image

Qubitium:
@avtc  Merge.  Our ci is broken right now. Please do a tiny small model test on `main`.  It doesn't even have to finish, just make sure it can quant like 1-3 layers without error. Need to make sure it has no regressions.

avtc:
@Qubitium 
I have successfully quantized 3 layers for qwen3b-30b-a3b with settings:
```
    offload_to_disk=False,
    wait_for_submodule_finalizers=True,
    gc_mode="on_stage_end",
    moe=MoEConfig(routing=ExpertsRoutingBypass()),
    auto_forward_data_parallel=False,
    calibration_data_device="balanced",
```
also checked with `calibration_data_device=None,` and with removed.
looks fine.

PR #2622, opened 2026-03-26
paro: clamp learned channel scales to avoid collapse

Qubitium:
@liang2kl Based on your suggestion, this PR now clamps the learned scales to avoid catastraphic collapse for high epoch/samples combo. Likely much more likely the samller the model. 

it does avoid the collapse to noise during inference. 

Not sure if this what you are thinking? Will do more tests.

### DIFF
--- a/tests/models/test_llama3_2_paroquant.py
+++ b/tests/models/test_llama3_2_paroquant.py
@@ -83,13 +83,13 @@ class TestLlama3_2_ParoQuant(ModelTest):
                 "floor_pct": 0.04,
             },
         },
-        "mmlu_stem": {
-            "chat_template": False,
-            "acc": {
-                "value": 0.3850301300348874,
-                "floor_pct": 0.04,
-            },
-        },
+        # "mmlu_stem": {
+        #     "chat_template": False,
+        #     "acc": {
+        #         "value": 0.3850301300348874,
+        #         "floor_pct": 0.04,
+        #     },
+        # },
     }
     FORMAT = FORMAT.PAROQUANT
     METHOD = METHOD.PAROQUANT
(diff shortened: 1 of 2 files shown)


tests/models/test_llama3_2_paroquant.py
```diff
@@ -94,9 +94,9 @@ class TestLlama3_2_ParoQuant(ModelTest):
     KERNEL_INFERENCE = {ParoQuantQuantLinear}
 
     # Mirror benchmark settings: 1+1 epochs on 128 train rows
-    PAROQUANT_ROTATION_EPOCHS = 5
-    PAROQUANT_FINETUNE_EPOCHS = 5
-    PAROQUANT_TRAIN_SAMPLES = 128
+    PAROQUANT_ROTATION_EPOCHS = 7
+    PAROQUANT_FINETUNE_EPOCHS = 7
+    PAROQUANT_TRAIN_SAMPLES = 1024
```
Qubitium: @liang2kl  here is the config change that would cause the inference collapse.

Qubitium:
epoch 7/7 + 1024 train sample now pass ci and score is now highest observed on avg vs previous but at a 5x time cost to optimization speed. 

```py
| Metric                                       |   PAROQUANT |
|----------------------------------------------|-------------|
| arc_challenge :: accuracy,loglikelihood      |      0.3191 |
| arc_challenge :: accuracy,loglikelihood_norm |      0.3524 |
| gsm8k_platinum_cot :: acc,num                |      0.4219 |
| mmlu_stem :: acc,ll                          |      0.4056 |
| mmlu_stem :: acc,ll_avg                      |      0.4056 |
```

Qubitium:
@liang2kl The clamp control works well but the clamp min/max is likely model specific and can be observed/learned. Using a default learned from the small llama 3.2 1b for now.  

based on my ci tests,  naive strategy to tune paroquant is `sample size` over `epoch`. Better to increase sample size and lower epoch than the reverse due to timing cost vs return on quality improvement.

Qubitium:
Latest test is not comparable with previoius as this `gsm8k_platinum` is `all` rows where as before it was limtied to 256 for faster ci.

paro, 2 epoch 2028 samples. 

```py
| Metric                                       |   PAROQUANT |
|----------------------------------------------|-------------|
| arc_challenge :: accuracy,loglikelihood      |      0.3234 |
| arc_challenge :: accuracy,loglikelihood_norm |      0.3558 |
| gsm8k_platinum_cot :: acc,num                |      0.4739 |
```

gptq 

```py
| Metric                                       |   MARLIN |
|----------------------------------------------|----------|
| arc_challenge :: accuracy,loglikelihood      |   0.3276 |
| arc_challenge :: accuracy,loglikelihood_norm |   0.3524 |
| gsm8k_platinum_cot :: acc,num                |   0.4425 |
```

PR #2727, opened 2026-04-14
[FIX] Mixtral MoE checkpoint module name may not match modeling code

ZX-ModelCloud:
## Summary

  Fix `LazyTurtle` so it can materialize legacy Mixtral-style MoE checkpoint tensors after
  runtime MoE blocks have been renamed by defusion.

  This covers cases where runtime names such as:

  - `model.layers.0.moe.experts.1.gate.weight`
  - `model.layers.0.moe.experts.1.up.weight`
  - `model.layers.0.moe.experts.1.down.weight`

  need to resolve back to legacy checkpoint keys such as:

  - `model.layers.0.block_sparse_moe.experts.1.w1.weight`
  - `model.layers.0.block_sparse_moe.experts.1.w3.weight`
  - `model.layers.0.block_sparse_moe.experts.1.w2.weight`


gptqmodel/utils/structure.py
```diff
@@ -565,6 +566,21 @@ class LazyTurtle:
...
+        "w2": ("w2", "down_proj", "down"),
+        "down_proj": ("down_proj", "down", "w2"),
+        "down": ("down", "down_proj", "w2"),
+        "w3": ("w3", "up_proj", "up"),
+        "up_proj": ("up_proj", "up", "w3"),
+        "up": ("up", "up_proj", "w3"),
+    }
+    _MOE_LEAF_PATTERN = re.compile(
+        r"^(?P<prefix>(?:.*\.)?experts\.\d+)\.(?P<leaf>w1|w2|w3|gate_proj|gate|up_proj|up|down_proj|down)\.(?P<tensor>weight|bias)$"
+    )
```
Qubitium: Try to get as much info from the model definition `module tree` instead of hard coding everything. We cannot avoid all hard code but at least reduce them.

Qubitium:
Doing refractor. This PR is doing WAY too much changes. It fixing one bug but injecting whole bunch of `aliasing` that is irrelevant to the Mixtral bug

PR #2741, opened 2026-04-15
[CI] mount ci host path as uv cache dir

CSY-ModelCloud:


CSY-ModelCloud:
mount /github/workspace/uv as uv cache dir

Qubitium (approved):

PR #2743, opened 2026-04-15
Fix Kernel JIT compat when torch cuda does not match local nvcc toolchain

groxaxo:
## Summary
- finish the Gemma 4 package/export cleanup and move the MiniMax HF smoke test into tests/
- update README/frontpage notes for Gemma 4 plus Ampere 3090/3060 coverage
- fix Marlin/Machete JIT flag gating to use the local nvcc version so Ampere hosts with older local toolchains do not fail on `-static-global-template-stub=false`

## Validation
- `pytest -q tests/test_model_definition_exports.py`
- `python tests/models/test_minimax_m2_hf.py --help`
- `CUDA_VISIBLE_DEVICES=2,3 GPTQMODEL_MARLIN_FORCE_REBUILD=1 python ...` loading `/home/op/outputs/gemma4-abliterated/gptq-pro-4bit` across both RTX 3060 GPUs with `MarlinLinear` selected and generation succeeding

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
 </p>
 
 ## Latest News
+* 04/15/2026 [main]: ✅Gemma 4 GPTQ package support completed end-to-end. `gptqmodel.models.definitions` now exports the full supported model-definition surface, and the Ampere CUDA path was revalidated on RTX 3090 / RTX 3060 (`sm_86`) hardware with a local Gemma 4 GPTQ model.
 * 04/08/2026 [main]: ✨ All CUDA kernels are now JIT compiled. The PyPI package is now about 300x smaller. On first use, GPT-QModel compiles only the kernels your workload actually needs. Improved Bonsai kernels now support execution `profile` control for `fast` or `low_memory` inference. Model weight loading during quantization has been optimized for large models like `GLM 5.1`. Added `GLM 5` and `GLM 5.1` model support.
 * 04/03/2026 [6.0.3](https://github.com/ModelCloud/GPTQModel/releases/tag/v6.0.3): 🎉 New quantization methods: `ParoQuant`, `GGUF`, `FP8`, `EXL3`, and `FOEM: First-Order Error Matters`. Added PrismML/Bonsai 1bit model quantization (inference only), faster ParoQuant/AWQ kernels, ParoQuant `optimization scope` control: `module` (Paro Lite) or `layer` (Paro reference), plus `Gemma4`, `MiniCPM-O`, `MiniCPM-V`, and `GLM4 MOE lite` model support.
 * 03/19/2026 [5.8.0](https://github.com/ModelCloud/GPTQModel/releases/tag/v5.8.0): ✨HF Transformers 5.3.0 support with auto-defusing of `fused` models via pypi pkg: [Defuser](https://github.com/ModelCloud/Defuser). Qwen 3.5 family support added. New fast HF `cpu` kernels for GPTQ/AWQ added. Experimental INT8 `cpu` kernel added for GPTQ. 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -278,7 +279,7 @@ GPT-QModel is validated for Linux, MacOS, and Windows 11:
 
 | Platform        | Device        |     |  Optimized Arch          | Kernels                                       |
 |-----------------|---------------| --- | ------------ |-----------------------------------------------| 
-| 🐧 Linux           | Nvidia GPU    | ✅       | `Ampere+` | Marlin, Exllama V2, Exllama V1, Triton, Torch |
+| 🐧 Linux           | Nvidia GPU    | ✅       | `Ampere+` (`sm_80`/`sm_86`, incl. RTX 3090 / 3060) | Marlin, Exllama V2, Exllama V1, Triton, Torch |
```
Qubitium, line 282: @groxaxo No need to list all the gpus for Ampere. This part can be removed or else the table is too wide for ux.

gptqmodel/looper/stage_subset.py
```diff
@@ -1030,11 +1030,17 @@ def _process_on_worker(
...
-        # parallel as allowed by the device thread pool.
+        # Use submit_serial for CUDA to avoid glibc pthread priority-inheritance
+        # assertion (Ubuntu 24.04 glibc 2.39) when concurrent CUDA threads run
+        # simultaneous Cholesky decompositions.
         tgt_dev = quant_target_devices.get(name, cur_layer_device)
+        _submitter = (
+            DEVICE_THREAD_POOL.submit_serial
+            if torch.device(tgt_dev).type in ("cuda", "xpu", "mps")
+            else DEVICE_THREAD_POOL.submit
+        )
```
Qubitium, line 1041: @groxaxo DId you hit this bug? I remember there is a Cholesky init bug that is not thread-safe but I mitigated with a thread safe `Cholesky` warm-up before first use. Please let me know if you are hitting this bug and if you can reproduce it in a `unit test`. If you can create a unit test that can recreate this bug under nogil Python, this will be most awesome.

Qubitium (changes requested):
@groxaxo Thank you for the 3090 runtime compat fix! Please check my comments. One is nitpick, ther other regarding threaded gptq calculations which may reduce performence for MoE quantization if we do not allow Cholesky for threads.

gptqmodel/quantization/gptq.py
```diff
@@ -966,6 +966,8 @@ def quantize(
         # H = self.H.to(device=self.H.device)
 
         if use_hessian:
+            # Replace NaN/Inf in H before processing (can occur with some model architectures)
+            self.H.nan_to_num_(nan=0.0, posinf=0.0, neginf=0.0)
```
Qubitium, line 970: Can you explain how you go this error and on which model? We do not want to hide/silent NaN errors as it causes quantization collapse for this block/module. Need to get to the bottom of this and not hide it under a rug.

gptqmodel/utils/marlin.py
```diff
@@ -195,7 +184,7 @@ def _marlin_extra_cuda_cflags() -> list[str]:
         include_fatbin_compression=True,
         include_diag_suppress=True,
     )
-    if _marlin_cuda_version_at_least(12, 8):
+    if local_nvcc_version_at_least(12, 8):
```
Qubitium, line 187: @groxaxo  This part/fix appears to show the Torch installed has mismatch with the `nvcc` (cuda-toolkit) package installed in os then it comes to support cuda version. is this corrrect? Pytorch has an old version but your local cuda-toolkit has a higher version? I am trying to figure the failure condition.

can you show me your torch version and torch exposed cuda version and your local nvcc version?

Qubitium:
@groxaxo Currently this PR contains things beyond what the title is fixing (torch version older than nvcc which actually compile the Marlin kernel) but the compiled kernel will be able to load find in torch and run: kind of like Torch is 12.6 but nvcc toolkit is 13.0 sort of system pkg mismatch, am I right to frame it like this?

Qubitium:
@groxaxo  I have fixed the actual nvcc compat check issue in https://github.com/ModelCloud/GPTQModel/pull/2745. Please let me know if you can provide unit tests that can show the other issues still exists and not resolved in your env. Thanks.

gptqmodel/looper/stage_subset.py
Qubitium: This `error` is likely caused by a threading bug that was fixed in https://github.com/ModelCloud/GPTQModel/pull/2749

Qubitium:
@groxaxo I am going to close this PR. The valid bug (nvcc + torch version conflict) has been fixed in another PR and the other issues in your PR but not in the summary has also been fixed, if applicable, at the src. Please try out `main`. 

The `NaN` issue is not a bug but modeling error due to MoE routing and/or you feeding it a little too little calibration data. We should have it crash, and not paper over it, so user know something is truly wrong. Hiding it would may actually cripple the quant task without any recovery. 

If your moe model is geting NaN, try enabling `fallback` and/or try full MoERouting strategy.

PR #2797, opened 2026-04-21
Fix AWQ JIT arch selection and add safe LLM-AWQ fallback

CSY-ModelCloud:
Problem: AWQ JIT builds could inherit unsupported CUDA arch targets such as sm_120 from the Torch wheel, and LLMAwqLinear fast kernels could still trigger illegal memory access on this mixed-GPU host during inference.

Fix: sanitize the effective TORCH_CUDA_ARCH_LIST against the local nvcc-supported targets before JIT compilation, add regression coverage for the nvcc fallback path, document the modified methods, and route GEMV_FAST/LLM_AWQ CUDA execution through a reference dequantize-plus-matmul path by default unless GPTQMODEL_AWQ_GEMV_FAST_FORCE_KERNEL=1 is set.

### DIFF
--- a/gptqmodel/nn_modules/qlinear/gemv_fast_awq.py
+++ b/gptqmodel/nn_modules/qlinear/gemv_fast_awq.py
@@ -17,7 +17,6 @@ from ...utils.awq import (
     awq_runtime_error,
 )
 from ...utils.backend import BACKEND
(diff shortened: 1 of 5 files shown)


CSY-ModelCloud (commented):
ready

PR #2825, opened 2026-04-26
Fix exllamav3_torch import under meta-device context

dblundell:
## Summary

Module import of `gptqmodel.nn_modules.exllamav3_torch` evaluates two
constants via `_half_scalar_from_bits(...)`, which previously called
`torch.tensor([bits], dtype=torch.uint16).view(torch.float16).item()`.
Under a `with torch.device('meta'):` preload context this raises:

```
RuntimeError: Tensor.item() cannot be called on meta tensors
```

This blocks any GPTQ load path on `transformers >= 5.6` that goes
through the AWQ preload pathway, because `replace_with_awq_linear`
imports `gptqmodel.quantization` (and transitively
`gptqmodel.nn_modules.exllamav3_torch`) inside a meta-device guard.

The fix replaces the tensor round-trip with a `struct`-based IEEE-754
binary16 conversion (`<e` format). It is bit-equivalent to the
tensor-based path and does not allocate a tensor, so it is unaffected
by the active device.

### Reproducer (without the fix)

```python
import torch
import sys

# Make sure we're forcing a fresh module body evaluation under meta.
for mod in list(sys.modules):
    if mod.startswith("gptqmodel.nn_modules.exllamav3_torch"):
        del sys.modules[mod]

with torch.device("meta"):
    import gptqmodel.nn_modules.exllamav3_torch  # RuntimeError on main
```

Encountered downstream while loading GPTQ-quantized Llama / Mistral /
Qwen checkpoints under `transformers 5.6.x` with `AutoModelForCausalLM`,
which preloads on `meta` device. With the patch applied the import
succeeds and the canonical EXL3 constants take their expected values.

## What Changed

- `gptqmodel/nn_modules/exllamav3_torch.py`: rewrite
  `_half_scalar_from_bits` to use `struct.pack('<H', ...)` +
  `struct.unpack('<e', ...)` instead of `torch.tensor(...).item()`. Add
  an explanatory comment that names the failure mode and the AWQ
  preload pathway that triggered it.
- `tests/test_exllamav3_meta_import.py`: new test module with three
  regression cases (see Tests).

No public API changes. The function signature and return values are
unchanged for every uint16 bit pattern (verified bit-equivalent across
representative cases including the two canonical EXL3 constants
`0x1EEE` / `0xC931` plus +0.0, +/-1.0, max/min finite, and NaN
patterns 0x7FFF / 0xFFFF).

## Tests

Every working PR must include at least one new simple, fast, targeted unit test when the change affects behavior, a bug fix, or a regression path.

The new file `tests/test_exllamav3_meta_import.py` adds three cases:

1. `test_half_scalar_from_bits_matches_torch_view_path` -
   bit-equivalence between the new `struct`-based path and the
   historical `torch.tensor(...).view(torch.float16).item()` path
   across +0.0, +/-1.0, max/min finite, the two canonical EXL3
   constants `0x1EEE` / `0xC931`, and NaN bit patterns.
2. `test_module_imports_under_meta_device` - the module loads cleanly
   inside `with torch.device('meta'):`. Reproduces the
   transformers AWQ pathway that surfaced the bug.
3. `test_module_constants_are_canonical_values` - pins
   `_EXL3_MUL1_INV` to `0.00676727294921875` and `_EXL3_MUL1_BIAS` to
   `-10.3828125`, the float64 IEEE-754 binary16 expansions of the
   canonical bit patterns.

Local validation - the conversion is verified bit-equivalent to the
tensor path on CPU across the test bit set, and the module imports
under `torch.device('meta')` without raising:

```bash
$ python -c "
import struct, torch
for bits in (0x1EEE, 0xC931, 0x0000, 0x3C00, 0xBC00, 0x7BFF, 0xFBFF):
    via_struct = float(struct.unpack('<e', struct.pack('<H', bits & 0xFFFF))[0])
    via_torch = float(torch.tensor([bits], dtype=torch.uint16).view(torch.float16).item())
    assert via_struct == via_torch, (bits, via_struct, via_torch)
print('bit-equivalence OK')

with torch.device('meta'):
    v = float(struct.unpack('<e', struct.pack('<H', 0x1EEE & 0xFFFF))[0])
    print('meta path OK:', v)
"
bit-equivalence OK
meta path OK: 0.00676727294921875
```

The full repo test suite was not exercised locally because it requires
a CUDA GPU and a populated HF cache; the targeted regression covers
the import path that this patch touches.

## Review Requirements

AI-assisted code is welcome.

## Notes

- The same struct-based formulation has been running in production in a
  downstream stack (`transformers 5.6.2 + gptqmodel 6.0.3 + torch 2.11`)
  applied as a post-install patch, across multiple GPTQ + AWQ load paths,
  with no observed numerical drift relative to the tensor-based original.
- `struct` is a stdlib module, so this introduces no new dependency.
- Endianness: `<e` is little-endian binary16 and `torch.float16`'s byte
  layout is also little-endian on all Tier 1 platforms (x86_64, aarch64).
  If a hypothetical big-endian build were ever supported, both the old
  tensor `.view(torch.float16)` path and the new `struct` path would
  need the same byte-order treatment, so this PR doesn't widen the
  endianness assumption that already exists in the file.

### DIFF
--- a/gptqmodel/nn_modules/exllamav3_torch.py
+++ b/gptqmodel/nn_modules/exllamav3_torch.py
@@ -7,6 +7,7 @@
 from __future__ import annotations
 
 import math
+import struct
 from functools import lru_cache
 from typing import Any, Dict, Optional
 
@@ -26,7 +27,20 @@ _EXL3_LOP3_BIAS = 0x3B603B60
 
 
 def _half_scalar_from_bits(bits: int) -> float:
-    return float(torch.tensor([bits], dtype=torch.uint16).view(torch.float16).item())
+    # Convert a uint16 bit pattern to its IEEE-754 binary16 (float16) value
+    # without allocating a torch tensor. The previous implementation used
+    # ``torch.tensor([bits], dtype=torch.uint16).view(torch.float16).item()``
+    # at module import, which fails under a meta-device preload context with
+    # ``RuntimeError: Tensor.item() cannot be called on meta tensors``. The
+    # transformers AWQ pathway (``replace_with_awq_linear``) imports
+    # ``gptqmodel.quantization`` inside ``with torch.device('meta'):``, which
+    # triggered the failure for any GPTQ load on transformers >= 5.6.
+    #
+    # ``struct`` format ``<e`` is IEEE-754 binary16 little-endian and matches
+    # ``torch.float16``'s byte layout on all supported platforms (x86_64 /
+    # aarch64). The conversion is bit-equivalent to the tensor-based path.
+    packed = struct.pack("<H", int(bits) & 0xFFFF)
+    return float(struct.unpack("<e", packed)[0])
 
 
 _EXL3_MUL1_INV = _half_scalar_from_bits(0x1EEE)
--- a/tests/test_exllamav3_meta_import.py
+++ b/tests/test_exllamav3_meta_import.py
@@ -1,0 +1,99 @@
+# SPDX-FileCopyrightText: 2026 ModelCloud.ai
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression test for module-level import of exllamav3_torch under a
+meta-device preload context.
+
+Background
+----------
+``gptqmodel.nn_modules.exllamav3_torch`` evaluates two module-level constants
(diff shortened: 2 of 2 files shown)


ZX-ModelCloud (approved):

Qubitium:
@dblundell Thank you for the contribution！ We have both verified the compat bug and this PR fix.

PR #2879, opened 2026-05-12
fix: stabilize MiniMax M2 meta materialization in quant replay

CSY-ModelCloud:


Qubitium:
@CSY-ModelCloud  We have some nasty monkey patching around this MiniMax M2 bug. Can we extract it into a base model property insead of checking for model names in our hf/structure classes?

PR #2921, opened 2026-06-11
[MODEL] support gemma4_unified

HaozheZhang6:
## Summary

`gemma4_unified` (multimodal Gemma 4, e.g. `OpenYourMind/gemma-4-12B-it-abliterated-uncensored`) fails with `Unsupport model_type gemma4_unified, and failed to auto-detect module tree`. It's a close sibling of the existing `gemma4` composite model, but its text stack drops the per-layer input adapters, so it needs its own definition rather than reusing `Gemma4ForConditionalGenerationGPTQ`.

## What Changed

- New `Gemma4UnifiedForConditionalGenerationGPTQ` definition over `model.language_model.layers`: per-projection q/k/v norms + dual pre/post feed-forward norms (same as `gemma4`), but **no** `per_layer_input_gate` / `post_per_layer_input_norm` / `per_layer_projection` — this variant has none. Its `language_model` also has no `project_per_layer_inputs`, which is why reusing the `gemma4` class `AttributeError`s during the calibration generate hook.
- `prepare_layer_replay_kwargs` refreshes the rope per layer: the model builds one `(cos, sin)` per attention `layer_type` and hands each decoder layer the tuple for its own type, so cached-layer replay has to regenerate it for the layer's `layer_type` (sliding vs full). Without it, replay reuses a stale tuple and hits a rope shape mismatch in `apply_rotary_pos_emb`.
- Registered `gemma4_unified` in `MODEL_MAP` and re-exported from `definitions/__init__.py`.

Vision/audio towers stay as base (non-quantized) modules, same as `gemma4`.

## Tests

`tests/test_gemma4_unified_support.py` (CPU-only): model_type → definition resolution, module tree excludes the per-layer-input paths, and the rope refresh returns the right tuple per `layer_type`.

```bash
$ pytest tests/test_gemma4_unified_support.py -v
tests/test_gemma4_unified_support.py::test_gemma4_unified_model_type_selects_definition PASSED
tests/test_gemma4_unified_support.py::test_gemma4_unified_module_tree_excludes_per_layer_input_paths PASSED
tests/test_gemma4_unified_support.py::test_gemma4_unified_replay_kwargs_refresh_position_embeddings_per_layer_type PASSED
3 passed in 3.93s
```

I also ran a full quantize → save → reload → generate on a tiny `gemma4_unified` (4-bit, group_size 128) on an A100: quantization completes across both the sliding and full attention layers, the saved model reloads with `TritonV2Linear` packed layers, and generation runs without error.

## Review Requirements

## Notes

The definition follows `gemma4.py`'s rope-replay approach but is kept as a separate class/file since the per-layer-input machinery there doesn't apply here. I didn't have the full 12B checkpoint on hand, so the end-to-end run used a tiny synthetic `gemma4_unified` — happy to validate against a specific checkpoint if you'd like.

### DIFF
--- a/gptqmodel/models/auto.py
+++ b/gptqmodel/models/auto.py
@@ -97,6 +97,7 @@ from .definitions.gemma2 import Gemma2QModel  # noqa: E402
 from .definitions.gemma3 import Gemma3ForConditionalGenerationGPTQ, Gemma3QModel  # noqa: E402
 from .definitions.gemma3n import Gemma3nForConditionalGenerationGPTQ, Gemma3nTextQModel  # noqa: E402
 from .definitions.gemma4 import Gemma4ForConditionalGenerationGPTQ, Gemma4TextQModel  # noqa: E402
+from .definitions.gemma4_unified import Gemma4UnifiedForConditionalGenerationGPTQ  # noqa: E402
 from .definitions.glm import GlmQModel  # noqa: E402
 from .definitions.glm4_moe import GLM4MoEGPTQ  # noqa: E402
 from .definitions.glm4_moe_lite import Glm4MoeLiteQModel  # noqa: E402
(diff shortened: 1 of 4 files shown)


Qubitium (approved):
@HaozheZhang6 Thank you! LGTM.

PR #2922, opened 2026-06-11
feat(models): register gemma4_unified GPTQ definition (#2920)

Anai-Guo:
## Summary

`gemma4_unified` checkpoints (`Gemma4UnifiedForConditionalGeneration`, e.g. `OpenYourMind/gemma-4-12B-it-abliterated-uncensored`) currently fail with:

```
ValueError: Unsupport model_type gemma4_unified, and failed to auto-detect module tree for model
```

This adds quantization definitions for the text-only (`gemma4_unified_text`) and composite (`gemma4_unified`) variants and registers them in `MODEL_MAP`.

Fixes #2920.

## Why a new definition instead of reusing `gemma4`

`gemma4_unified` is **not** a drop-in alias for `gemma4`:

- `Gemma4UnifiedTextConfig` explicitly removes the per-layer-input (PLE) feature (`vocab_size_per_layer_input` / `hidden_size_per_layer_input` set to `AttributeError()`). So the decoder has **no** `per_layer_input_gate` / `per_layer_projection` / `post_per_layer_input_norm` modules and **no** `project_per_layer_inputs` hook.
- Reusing `Gemma4ForConditionalGenerationGPTQ` would therefore crash in `pre_quantize_generate_hook_start` → `_patch_gemma4_per_layer_input_capture` (no `project_per_layer_inputs`), and its positional-input capture assumes a `per_layer_input` arg the unified decoder does not have.

The quantizable layout is the standard Gemma sandwich-norm decoder (`q/k/v/o_proj` + `gate/up/down_proj`, with `q/k/v_norm` and the four sandwich norms non-quantized).

`gemma4_unified` **does** keep per-layer-type rotary frequencies (sliding vs full attention via `Gemma4UnifiedTextRotaryEmbedding`), so cached replay reuses the proven `_prepare_gemma4_replay_kwargs` helper to refresh `position_embeddings` per layer — the PLE-reconstruction tail in that helper is an inert no-op here since no per-layer inputs are captured.

## Test plan

- [ ] `from gptqmodel.models import auto; auto.check_and_get_model_definition(<gemma4_unified ckpt>)` returns `Gemma4UnifiedForConditionalGenerationGPTQ`.
- [ ] End-to-end 4-bit quantization of a `gemma4_unified` checkpoint.

I don't have local access to a `gemma4_unified` checkpoint to run the full quantization, so maintainer validation on real hardware is appreciated. The definition deliberately mirrors the existing (tested) `gemma4.py` pattern.

[...]

PR #2925, opened 2026-06-17
feat(models): register gemma4_unified_text text-only GPTQ definition

Anai-Guo:
## Summary

Follow-up to #2921 (which added the composite `gemma4_unified`). The `gemma4_unified` family also exposes a **standalone text stack** — `model_type = "gemma4_unified_text"` (`Gemma4UnifiedTextModel`), registered in `transformers` (`src/transformers/models/auto/auto_mappings.py` / `modeling_auto.py`) — exactly mirroring how `gemma4_text` parallels `gemma4`.

`MODEL_MAP` currently has `gemma4_text`, `gemma4`, and `gemma4_unified` but **not** `gemma4_unified_text`, so a text-only unified checkpoint still fails with:

```
ValueError: Unsupport model_type gemma4_unified_text, and failed to auto-detect module tree for model
```

## What this does

- Extracts the shared Gemma 4 unified decoder layout into `_GEMMA4_UNIFIED_DECODER` (no behavior change for the composite definition).
- Adds `Gemma4UnifiedTextQModel`, rooted at `model.layers` with `model.norm` / `model.rotary_emb` (the standalone text-model layout, no `language_model` nesting), reusing the same per-layer-type rope replay helper as the composite variant.
- Registers `"gemma4_unified_text"` in `MODEL_MAP`.

The quantizable layout is identical to the composite (`q/k/v/o_proj` + `gate/up/down_proj`, with `q/k/v_norm` and the four sandwich norms non-quantized); only the module-tree root and norm/rotary paths differ, just as `Gemma4TextQModel` differs from `Gemma4ForConditionalGenerationGPTQ`.

## Test plan

- [ ] `auto.check_and_get_model_definition(<gemma4_unified_text ckpt>)` returns `Gemma4UnifiedTextQModel`.
- [ ] End-to-end 4-bit quantization of a text-only `gemma4_unified_text` checkpoint.

I don't have local access to a `gemma4_unified_text` checkpoint to run the full quantization, so maintainer validation on real hardware is appreciated. The definition deliberately mirrors the merged (tested) composite `gemma4_unified` pattern.

cc @Qubitium — this is the "additional patching" for the text-only variant you asked about in #2922.

### DIFF
--- a/gptqmodel/models/auto.py
+++ b/gptqmodel/models/auto.py
@@ -97,7 +97,7 @@ from .definitions.gemma2 import Gemma2QModel  # noqa: E402
 from .definitions.gemma3 import Gemma3ForConditionalGenerationGPTQ, Gemma3QModel  # noqa: E402
 from .definitions.gemma3n import Gemma3nForConditionalGenerationGPTQ, Gemma3nTextQModel  # noqa: E402
 from .definitions.gemma4 import Gemma4ForConditionalGenerationGPTQ, Gemma4TextQModel  # noqa: E402
-from .definitions.gemma4_unified import Gemma4UnifiedForConditionalGenerationGPTQ  # noqa: E402
+from .definitions.gemma4_unified import Gemma4UnifiedForConditionalGenerationGPTQ, Gemma4UnifiedTextQModel  # noqa: E402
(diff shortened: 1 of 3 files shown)


Qubitium:
@Anai-Guo LGTM. Thank you!

PR #2926, opened 2026-06-18
Fix Mllama (Llama3.2-11B-Vision) quantization

JeevanBhoot:
## Summary

Fix Mllama quantization for current Transformers Mllama layouts

Bugs:
- Mllama definition in GPTQModel pointed at the old decoder path `language_model.model.layers`, but this is now exposed as `model.language_model.layers` in Transformers, so no layers were found for quantization
- LazyTurtle shell names did not match checkpoint names
- Calibration input capture ran through the full HF multimodal wrapper, which builds masks and can call `.item()` on meta tensors.
- Some calibration tensors could end up on a different device from the embedding module.

## What Changed

- Update `MLlamaQModel` to find decoder layers under the current `model.language_model.layers` layout while keeping the legacy `language_model.model.layers` path.
- Add an Mllama LazyTurtle conversion map for runtime shell paths such as `model.language_model.*` to checkpoint paths such as `language_model.model.*`.
- Materialize `embed_tokens` and `rotary_emb` before calibration input capture, then move/offload them back after capture.
- Override Mllama input capture to call the first language decoder layer directly with computed embeddings, position ids, rotary embeddings, and a normalized all-true attention mask.
- Add fast unit coverage for Mllama layout resolution, base-module discovery, LazyTurtle path conversion, input-capture routing, and input-id device handling.

## Tests

Every working PR must include at least one new simple, fast, targeted unit test when the change affects behavior, a bug fix, or a regression path.

Paste the exact test commands and results here:

```bash
python -m pytest tests/test_mllama_support.py -q
# 6 passed, 14 warnings in 1.95s
```

## Review Requirements

AI-assisted code is welcome.

### DIFF
--- a/gptqmodel/models/definitions/mllama.py
+++ b/gptqmodel/models/definitions/mllama.py
@@ -20,6 +20,8 @@ class MLlamaQModel(BaseQModel):
     loader = AutoModelForPreTraining
 
     pre_lm_head_norm_module = "model.language_model.norm"
+    # Current Transformers shells use `model.language_model.*`, while the
+    # released Mllama checkpoints store those tensors under `language_model.model.*`.
     HF_CONVERSION_MAP_REVERSED = (
         SimpleNamespace(
             source_patterns=[r"model\.language_model"],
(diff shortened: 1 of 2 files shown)


Qubitium (approved):

Qubitium:
@JeevanBhoot Thank you for the root model prefix compat fix.

PR #2931, opened 2026-06-29
Downgrade defuser to 0.0.22

JeevanBhoot:
Can not currently install GPTQModel from main because defuser 0.0.23 does not exist.


Qubitium:
@JeevanBhoot  Defuser 0.0.23 has been released.

PR #2943, opened 2026-07-07
Fix ExLlamaV2 kernels on Windows: link cublas.lib and enable PLATFORM.WIN32

### LINKED ISSUE #2936: [Windows] ExLlamaV2 GPTQ/AWQ JIT build silently fails at link: unresolved external symbol cublasHgemm (missing cublas.lib in extra_ldflags)
## Summary

On Windows, the ExLlamaV2 JIT extensions (both GPTQ and AWQ) always fail to build at the final
link step because `cublas.lib` is never passed to the MSVC linker. The failure is **silent**:
gptqmodel logs `torch.ops JIT compilation failed ... using fallback path` and quietly selects a
slower kernel, so users with a fully working MSVC + CUDA toolchain still never get ExLlamaV2.

The newer ExLlamaV3 extension already handles this correctly via `_extra_ldflags()` in
`gptqmodel/exllamav3/ext.py` — the ExLlamaV2 extension definitions in
`gptqmodel/utils/exllamav2.py` simply don't pass any `extra_ldflags`.

## Environment

- gptqmodel 7.1.0 (PyPI), transformers 5.12.1, torch 2.6.0+cu124
- Windows 11, Python 3.10.6
- MSVC 14.39.33519 (VS 2022 Community, C++ workload), Windows SDK 10.0.22621.0
- CUDA toolkit 12.4.131 (nvcc), GPU: RTX 3050 Ti Laptop (SM 8.6)

Confirmed still present on current `main` (`gptqmodel/utils/exllamav2.py` has no
`extra_ldflags` / `cublas` reference as of 2026-07-04).

## Reproduction

1. Windows machine with working `cl.exe` + `nvcc` on PATH (`torch.utils.cpp_extension.load_inline`
   with a trivial CUDA kernel compiles fine).
2. Load any AWQ or GPTQ checkpoint through transformers (e.g.
   `Qwen/Qwen2-VL-2B-Instruct-AWQ`), or anything that triggers the ExLlamaV2 extension build.
3. Compilation of all `.cu`/`.cpp` objects succeeds, then linking fails.

Running `ninja -v` manually inside the build dir shows the actual error (gptqmodel swallows it):

```
[1/1] link.exe ext_awq.o q_matrix_awq.cuda.o q_gemm_awq.cuda.o /nologo /DLL c10.lib c10_cuda.lib
      torch_cpu.lib torch_cuda.lib ... cudart.lib /out:gptqmodel_exllamav2_awq_ops.pyd
q_gemm_awq.cuda.o : error LNK2019: unresolved external symbol cublasHgemm referenced in function
      "void __cdecl gemm_half_q_half_cuda(struct cublasContext *, ...)"
gptqmodel_exllamav2_awq_ops.pyd : fatal error LNK1120: 1 unresolved externals
```

What the user sees instead is only:

```
INFO  ExLlamaV2 AWQ: torch.ops JIT compilation failed in 57s (estimated ~35s); using fallback path.
```

## Root cause

`q_gemm.cu` / `q_gemm_awq.cu` call `cublasHgemm`, but neither
`_EXLLAMAV2_GPTQ_TORCH_OPS_EXTENSION` nor `_EXLLAMAV2_AWQ_TORCH_OPS_EXTENSION` in
`gptqmodel/utils/exllamav2.py` set `extra_ldflags`, so on Windows the MSVC link line never
includes `cublas.lib`. On Linux this happens to work because of how the shared library resolves
symbols at load time; MSVC requires the import library at link time.

## Suggested fix

Mirror `_extra_ldflags()` from `gptqmodel/exllamav3/ext.py`:

```python
def _exllamav2_ldflags() -> list[str]:
    if sys.platform != "win32":
        return []
    flags = ["cublas.lib"]
    if sys.base_prefix != sys.prefix:  # venv: python .lib lives in the base install
        flags.append(f"/LIBPATH:{os.path.join(sys.base_prefix, 'libs')}")
    return flags
```

and pass `extra_ldflags=_exllamav2_ldflags` to both `TorchOpsJitExtension(...)` definitions.

Verified locally: with this one-line wiring both extensions build and load
(`ExLlamaV2 AWQ: torch.ops JIT extension ready in 61s`), and inference output is correct.

A secondary suggestion: when a JIT build fails, logging the last few lines of the ninja error
(or pointing at the build dir) would make this class of problem much easier to diagnose —
the current single INFO line hides a hard toolchain error behind "using fallback path".

### LINKED ISSUE #2937: [Windows] ExLlamaV2 kernels are gated to Linux, but compile and produce correct output on Windows once linking is fixed
## Summary

`ExllamaV2Linear` and `AwqExllamaV2Linear` declare `SUPPORTS_PLATFORM = [PLATFORM.LINUX]`, so on
Windows they are excluded from kernel auto-selection, and forcing `backend="exllama_v2"` raises:

```
ValueError: <class 'gptqmodel.nn_modules.qlinear.exllamav2_awq.AwqExllamaV2Linear'>
does not support platform: win32
```

After fixing the Windows link failure (#2936) and locally changing the gate to
`[PLATFORM.LINUX, PLATFORM.WIN32]`, both kernels build, load, and produce correct output on
Windows — and they are the fastest kernels available there.

## Measurements

Qwen2-VL-2B-Instruct, RTX 3050 Ti Laptop 4 GB (SM 8.6), Windows 11, torch 2.6.0+cu124,
gptqmodel 7.1.0, greedy decoding, mean over a fixed 3-image document-understanding set with
keyword-recall scoring against ground truth:

| Checkpoint | Kernel | Decode tok/s | s/case | Accuracy |
|---|---|---|---|---|
| GPTQ-Int4 | TritonV2 (auto-selected) | 5.3 | 14.1 | 94% |
| GPTQ-Int4 | **ExLlamaV2 (forced)** | **7.6** | **9.6** | **94%** |
| AWQ | AwqGEMMTriton (auto-selected) | 6.1 | 10.4 | 78% |
| AWQ | AwqGEMM (compiled native) | 4.9 | 13.3 | 78% |
| AWQ | **ExLlamaV2 (forced)** | **6.7** | **9.1** | **78%** |

Accuracy is identical to the other kernels in every case (same per-case keyword hits/misses),
so the numerics look correct on win32.

## Ask

Was the Linux gate a deliberate correctness decision, or just "untested on Windows"? If the
latter, would you accept a PR enabling `PLATFORM.WIN32` for the ExLlamaV2 kernels (possibly
behind an env flag until CI coverage exists)? Happy to share more benchmark details.

xXpeira12:
Fixes #2936
Fixes #2937

As invited in both issues โ€” this makes the ExLlamaV2 GPTQ/AWQ kernels usable on Windows.

## Changes

1. **`gptqmodel/utils/exllamav2.py`** โ€” add `_exllamav2_extra_ldflags()` and wire it into both
   `TorchOpsJitExtension` definitions. `q_gemm.cu` / `q_gemm_awq.cu` call `cublasHgemm`; on Linux
   the symbol resolves at load time, but MSVC requires the import library at link time, so the
   Windows JIT build always failed with `LNK2019: unresolved external symbol cublasHgemm` and
   silently fell back to slower kernels. This mirrors `_extra_ldflags()` in
   `gptqmodel/exllamav3/ext.py`, which already handles this. No behavior change on non-Windows
   (returns `[]`).

2. **`gptqmodel/nn_modules/qlinear/exllamav2.py`**, **`exllamav2_awq.py`** โ€” add
   `PLATFORM.WIN32` to `SUPPORTS_PLATFORM` for both kernels.

## Testing

Windows 11, Python 3.10.6, torch 2.6.0+cu124, transformers 5.12.1, MSVC 14.39 (VS 2022),
CUDA toolkit 12.4.131, RTX 3050 Ti Laptop 4 GB (SM 8.6):

- Both extensions JIT-compile and load: `ExLlamaV2 GPTQ: torch.ops JIT extension ready in 48s`,
  `ExLlamaV2 AWQ: torch.ops JIT extension ready in 61s`.
- End-to-end inference with `Qwen/Qwen2-VL-2B-Instruct-GPTQ-Int4` and `-AWQ` through
  transformers, greedy decoding over a fixed 3-image document-understanding set with
  keyword-recall scoring:

| Checkpoint | Kernel | Decode tok/s | Accuracy |
|---|---|---|---|
| GPTQ-Int4 | TritonV2 (previous fallback) | 5.3 | 94% |
| GPTQ-Int4 | **ExLlamaV2 (this PR)** | **7.6** | **94%** |
| AWQ | AwqGEMMTriton (previous fallback) | 6.1 | 78% |
| AWQ | **ExLlamaV2 (this PR)** | **6.7** | 78% |

Accuracy and per-case keyword hits/misses are identical to the Triton kernels, so the numerics
match on win32.

I understand CI has no Windows coverage (#2937) โ€” happy to adjust if you'd prefer the platform
enablement gated behind an env flag instead.

### DIFF
--- a/gptqmodel/nn_modules/qlinear/exllamav2.py
+++ b/gptqmodel/nn_modules/qlinear/exllamav2.py
@@ -49,7 +49,9 @@ class ExllamaV2Linear(GPTQQuantLinear):
     SUPPORTS_OUT_FEATURES_DIVISIBLE_BY = [32]
 
     SUPPORTS_DEVICES = [DEVICE.CUDA] # ROCm has broken accuracies issues
-    SUPPORTS_PLATFORM = [PLATFORM.LINUX]
+    # Windows verified: kernel JIT-compiles with MSVC + CUDA 12.4 and produces
+    # output identical to the Triton kernel (benchmarks in #2937).
+    SUPPORTS_PLATFORM = [PLATFORM.LINUX, PLATFORM.WIN32]
     SUPPORTS_PACK_DTYPES = [torch.int32]
     SUPPORTS_ADAPTERS = [Lora]
 
--- a/gptqmodel/nn_modules/qlinear/exllamav2_awq.py
+++ b/gptqmodel/nn_modules/qlinear/exllamav2_awq.py
@@ -41,7 +41,9 @@ class AwqExllamaV2Linear(AWQuantLinear):
     SUPPORTS_OUT_FEATURES_DIVISIBLE_BY = [32]
 
     SUPPORTS_DEVICES = [DEVICE.CUDA] # ROCm has broken accuracies issues
-    SUPPORTS_PLATFORM = [PLATFORM.LINUX]
+    # Windows verified: kernel JIT-compiles with MSVC + CUDA 12.4 and produces
+    # output identical to the Triton kernels (benchmarks in #2937).
+    SUPPORTS_PLATFORM = [PLATFORM.LINUX, PLATFORM.WIN32]
     SUPPORTS_PACK_DTYPES = [torch.int32]
     SUPPORTS_ADAPTERS = [Lora]
 
--- a/gptqmodel/utils/exllamav2.py
+++ b/gptqmodel/utils/exllamav2.py
@@ -5,6 +5,8 @@
 
 from __future__ import annotations
 
+import os
+import sys
 from pathlib import Path
 
 import torch
@@ -93,6 +95,19 @@ def _exllamav2_extra_cuda_cflags() -> list[str]:
     )
 
 
+def _exllamav2_extra_ldflags() -> list[str]:
+    # q_gemm.cu / q_gemm_awq.cu call cublasHgemm. On Linux the symbol resolves
+    # at load time, but the MSVC linker needs the import library explicitly or
+    # the build fails with `LNK2019: unresolved external symbol cublasHgemm`.
+    # Mirrors _extra_ldflags in gptqmodel/exllamav3/ext.py.
+    if sys.platform != "win32":
+        return []
+    flags = ["cublas.lib"]
+    if sys.base_prefix != sys.prefix:
+        flags.append(f"/LIBPATH:{os.path.join(sys.base_prefix, 'libs')}")
+    return flags
+
+
 def _exllamav2_awq_extra_cflags() -> list[str]:
     return default_jit_cflags(opt_level=None, enable_bf16=True)
 
@@ -120,6 +135,7 @@ _EXLLAMAV2_GPTQ_TORCH_OPS_EXTENSION = TorchOpsJitExtension(
     extra_cflags=_exllamav2_gptq_extra_cflags,
     extra_cuda_cflags=_exllamav2_gptq_extra_cuda_cflags,
     extra_include_paths=_exllamav2_include_paths,
+    extra_ldflags=_exllamav2_extra_ldflags,
     force_rebuild_env="GPTQMODEL_EXLLAMAV2_FORCE_REBUILD",
     verbose_env="GPTQMODEL_EXT_VERBOSE",
     requires_cuda=True,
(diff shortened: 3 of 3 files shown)


xXpeira12:
Closing this one — will follow up with a reworked PR.

PR #2945, opened 2026-07-08
Fix O(N^2) state-dict collection that stalled saves of large MoE models for offload_to_disk=True

avtc:
## Summary

Saving a quantized model with `offload_to_disk=True` could hang for hours - one core at 100% CPU, near-zero disk I/O, no shards written.

The cause was an O(N²) bug in state-dict collection: `_non_persistent_buffers_set` was checked via `get_module_by_name()`, which walks the entire module tree for *each* buffer.

## What Changed

- **`gptqmodel/utils/model.py`** - Collect persistent buffers in a single `named_modules()` walk, reading each module's own buffers via `named_buffers(recurse=False)` and checking persistence inline. Fixed in both the offload and non-offload collection branches. Output state dict is identical.
- No API or behavior changes.

## Tests

```bash
pytest tests/test_save_state_dict_buffer_collection.py -v
tests/test_save_state_dict_buffer_collection.py::test_non_persistent_buffers_are_excluded PASSED
tests/test_save_state_dict_buffer_collection.py::test_all_persistent_parameters_and_buffers_present PASSED
tests/test_save_state_dict_buffer_collection.py::test_mixed_persistent_and_non_persistent_buffers PASSED
tests/test_save_state_dict_buffer_collection.py::test_nested_empty_module_prefix_is_safe PASSED
tests/test_save_state_dict_buffer_collection.py::test_collection_scales_linearly_with_module_count PASSED

pytest tests/test_offload_files.py -v        # passed - directly exercises changed code
pytest tests/test_out_of_model_tensors.py -v # passed - save-path sharding
```

## Review Requirements

## Notes

- After fix: For HY3 (~46K modules) - it took 13 minutes to write first safetensors after quantize_config was saved.
- Before fix: For MiniMax 2.7 - it took 3 hours and 10 minutes to write first safetensors file after quantize_config was saved.
- Buffer ordering follows `named_modules()` recursion (same traversal order `named_buffers()` already used), so shard planning is unaffected.

### DIFF
--- a/gptqmodel/utils/model.py
+++ b/gptqmodel/utils/model.py
@@ -1627,11 +1627,9 @@ def _collect_state_dict_with_offload(model: nn.Module, offload_root: str) -> Dic
             source = param
         state_dict[name] = TensorSource(name=name, torch_dtype=param.dtype, shape=tuple(param.shape), source=source)
 
-    # Walk the module tree ONCE and collect persistent buffers directly from each
-    # module. This avoids the previous O(N^2) behavior where every buffer triggered
-    # a full `get_module_by_name()` -> `named_modules()` traversal of the entire
-    # model (N buffers x N modules), which made saving large MoE models (e.g. 46K
-    # modules / ~200K buffers) take hours of pure-Python CPU work.
+    # Collect persistent buffers in a single module-tree walk: each module owns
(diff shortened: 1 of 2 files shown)


ZX-ModelCloud (approved):

Qubitium:
@avtc Awesome!

PR #2950, opened 2026-07-14
fix(triton): bound-check checkpoint g_idx to prevent OOB dequant read (#2949)

### LINKED ISSUE #2949: Out-of-bounds read in the Triton dequant kernel via an unchecked checkpoint `g_idx` (CWE-125)
**Summary.** Loading and running inference on a crafted GPTQ checkpoint (`desc_act=True`) can trigger an out-of-bounds GPU memory read in GPTQModel's Triton dequant kernel. The per-row group index `g_idx` is loaded from the checkpoint and used to index the `scales`/`qzeros` buffers with no check that its values are within `[0, num_groups)`, where `num_groups` is itself taken from the (attacker-supplied) `scales` tensor. A `g_idx` entry `>= num_groups` reads out of bounds, crashing the CUDA context (denial of service) and potentially exposing adjacent GPU memory. This requires loading an untrusted checkpoint; it is not reachable with a legitimate model.

**Details.** GPTQ checkpoints with activation reordering (`desc_act=True`) ship a `g_idx` tensor of shape `(in_features,)`. On load, `from_quantized` → `accelerate.load_checkpoint_in_model` → `set_module_tensor_to_device` replaces the layer's config-derived buffers with the checkpoint tensors and does not re-validate them; the Triton layer's `post_init` has its `g_idx` regeneration commented out, so the file's `g_idx` values reach the kernel verbatim. The Triton `dequant_kernel` (`gptqmodel/nn_modules/triton_utils/dequant.py`) then computes:

```python
g_idx  = tl.load(g_idx_ptr + row_idx, ...)
groups = tl.where(g_idx < 0, g_idx + num_groups, g_idx)   # only the negative-wrap; no upper bound
scales = tl.load(scales_ptr + (col_idx + out_features * groups), ...)   # no groups < num_groups check
qzeros = tl.load(qzeros_ptr + (qzero_ncols * groups + col_idx // pack_scale), ...)
```

with `num_groups = scales.shape[0]`, `out_features = scales.shape[1]`, `in_features = g_idx.shape[0]` — all read from the checkpoint. There is no `groups < num_groups` bound, so an out-of-range `g_idx` value indexes `scales`/`qzeros` past their extent. The `qzeros` load has the same unchecked `groups` factor. The safe Torch backend (`nn_modules/qlinear/torch.py`) performs the same gather with bounds-checked PyTorch indexing (`scales_tile[g_idx]`), which faults safely — only the Triton path reads raw.

**Proof of concept.** Confirmed on GPTQModel 7.1.0, torch 2.13.0+cu130, triton 3.7.1, NVIDIA A100, under `compute-sanitizer --tool memcheck` with `PYTORCH_NO_CUDA_MEMORY_CACHING=1`. Build a 4-bit `desc_act=True` GPTQ layer (in=256, out=128, group_size=128, so `num_groups=2`), load a `g_idx` tensor of the correct shape `(256,)` but with one out-of-range value (e.g. `g_idx[0] = 6`) via `accelerate.set_module_tensor_to_device` (the real load primitive accepts it with no bounds check), and run `forward()` on CUDA through the Triton dequant path. The valid case runs clean (0 errors); the crafted case reports an out-of-bounds read inside `dequant_kernel`:

```
Invalid __global__ read of size 4 bytes
    at dequant_kernel ... in dequant.py
    Access to 0x... is out of bounds
ERROR SUMMARY: 20 errors   (+ cudaErrorLaunchFailure)
```

On CPU, the Torch backend raises `IndexError: index 6 is out of bounds for dimension 0 with size 2` on the same input, confirming the value is genuinely out of range and that only the Triton path reads unchecked.

**Impact.** An attacker who can get a victim to load and run a crafted GPTQ checkpoint (a model file, e.g. from a model hub) can crash the inference process (CUDA-context launch failure = denial of service) and potentially read adjacent GPU memory into the dequantized weights.

**Suggested fix.** Validate `g_idx` values against `num_groups` (clamp or reject `g_idx >= num_groups`) at load (`post_init`), and/or add a `groups < num_groups` guard in the Triton kernel, matching the safe behavior of the Torch backend.

**Severity.** Medium — CVSS:3.1/AV:L/AC:L/PR:N/UI:R/S:U/C:L/I:N/A:H = 6.1. CWE-125 (out-of-bounds read). Affected: `gptqmodel <= 7.1.0` and current `main`.

---
Reported by Halo Forge Labs.

Anai-Guo:
## Summary

Fixes #2949 — an out-of-bounds device read (CWE-125) in the Triton dequant path when loading an untrusted GPTQ checkpoint.

The Triton dequant kernel (`gptqmodel/nn_modules/triton_utils/dequant.py`) resolves each row's group as `groups = where(g_idx < 0, g_idx + num_groups, g_idx)` and then indexes `scales`/`qzeros` with `groups` **with no upper-bound check**, where `num_groups = scales.shape[0]`. Because `TritonV2Linear.post_init` leaves g_idx regeneration disabled, a crafted `desc_act=True` checkpoint whose `g_idx` exceeds `num_groups` reaches the kernel verbatim and reads past the buffer — crashing the CUDA context (DoS) and potentially pulling adjacent GPU memory into the dequantized weights. The Torch backend's `scales[g_idx]` gather faults safely on the same input; only the Triton path reads raw.

## Fix

Validate `g_idx` once at load in `TritonV2Linear.post_init`, rejecting values outside `[-num_groups, num_groups)`. That range is **exactly** what the kernel's negative-wrap can resolve safely and what the Torch backend's `scales[g_idx]` gather accepts, so the guard only rejects inputs that already fault safely on the Torch path — no legitimate checkpoint is affected. It runs once at load, not on the per-forward hot path.

I kept the change to the load-side guard alone (rather than also masking the load inside the kernel) because rejecting at load fully prevents the bad `g_idx` from ever reaching the device; a kernel-side `groups < num_groups` mask would be reasonable defense-in-depth if you prefer it.

## Tests

`tests/test_triton_g_idx_bounds.py` — CPU-only unit tests for the pure validator (no GPU / Triton needed): in-range and negative-wrap-in-range pass; `g_idx >= num_groups` and `g_idx < -num_groups` are rejected; `None`/empty buffers are no-ops. The accept/reject boundaries were cross-checked against `torch.zeros((num_groups, n))[g_idx]` — they match the Torch backend's safe-fault semantics exactly.

### DIFF
--- a/gptqmodel/nn_modules/qlinear/tritonv2.py
+++ b/gptqmodel/nn_modules/qlinear/tritonv2.py
@@ -19,6 +19,37 @@ from .torch import TorchLinear
 log = setup_logger()
 
 
+def _validate_g_idx_bounds(
+    g_idx: Optional[torch.Tensor],
+    scales: Optional[torch.Tensor],
+    layer_name: str = "TritonV2Linear",
+) -> None:
+    """Reject checkpoint ``g_idx`` that would index the group buffers out of bounds.
+
+    The Triton dequant kernel resolves each row's group as
+    ``groups = where(g_idx < 0, g_idx + num_groups, g_idx)`` and then loads
+    ``scales``/``qzeros`` at that group with no upper-bound guard, where
+    ``num_groups = scales.shape[0]``. A value outside ``[-num_groups, num_groups)``
+    therefore indexes past the buffer, causing an out-of-bounds device read
+    (denial of service, possible adjacent-memory disclosure) when an untrusted
+    checkpoint is loaded. This mirrors the safe faulting of the Torch backend's
+    ``scales[g_idx]`` gather, but validates once at load rather than per forward.
+    """
+    if g_idx is None or scales is None or g_idx.numel() == 0:
+        return
+
+    num_groups = scales.shape[0]
+    g_min = int(g_idx.min())
+    g_max = int(g_idx.max())
+    if g_min < -num_groups or g_max >= num_groups:
+        raise ValueError(
+            f"{layer_name}: checkpoint g_idx is out of range for {num_groups} "
+            f"scale group(s) (got min={g_min}, max={g_max}, valid "
+            f"[{-num_groups}, {num_groups - 1}]); the group index would read "
+            f"scales/qzeros out of bounds."
+        )
+
+
 class TritonV2Linear(TorchLinear):
     SUPPORTS_BACKENDS = [BACKEND.GPTQ_TRITON]
     SUPPORTS_METHODS = [METHOD.GPTQ]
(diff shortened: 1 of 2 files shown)


Qubitium (approved):

Qubitium:
@Anai-Guo Thank you!

PR #3014, opened 2026-08-13
fix(eora): clamp covariance eigenvalues before inversion to avoid outlier adapters

Qubitium:
@nbasyl @hutm @cmhungsteve Please review. The following bug was encountered by me doing testing on EoRA. The lower the bpw, the higher the chance this becomes and the more damage it will cause. Most of the time it just manifest it self silently as lower EoRA recovery scores and was very sneaky to catch. It often does not manifest as catastrophic failure which is the sneaky part. 

## Summary

`eora_compute_lora` eigendecomposes the activation covariance matrix and forms a pseudoinverse via `1 / sqrt(eigenvalue)`. Numerical noise in the float32 accumulated covariance can yield tiny or slightly negative eigenvalues; the old code replaced negatives with the smallest positive eigenvalue, then computed `1/sqrt(lambda)` for every mode. When `lambda` is near zero this creates arbitrarily large inverse values that propagate as `NaN`/`Inf` outliers in the low-rank `A`/`B` tensors and, in the worst case, the branch `torch.min(L[L > 0])` crashes when every eigenvalue is non-positive.

Clamp the spectrum to a numerical-rank tolerance: discard eigenvalues at or below `max(lambda) * max(1, n) * float32_eps` by setting their `sqrt`/`rsqrt` contributions to zero, so they do not participate in the pseudoinverse. This produces finite adapters for rank-deficient or numerically indefinite covariances.

## What Changed

- `gptqmodel/eora/eora.py`: In `eora_compute_lora`, replace the `minimum`-replacement logic with a relative tolerance cutoff and zero-pad the discarded spectrum before `sqrt`/`rsqrt`. Also raise `FloatingPointError` if the eigensolve returns non-finite eigenvalues.
- `tests/test_eora.py`: New focused unit tests for covariance spectra containing negative/tiny eigenvalues and the all-negative pathological case.

## Tests

```bash
cd tests && python -m pytest test_eora.py -x -q
```

Result: `5 passed in 2.55s`

## Review Requirements

- [ ] Every changed file has been reviewed by a human.

## Notes

This fix keeps the existing `eora_compute_lora` signature and does not change the public EoRA API. The cutoff follows the same float32 numerical-rank reasoning used in the codebase's other Hessian/covariance paths.

### DIFF
--- a/gptqmodel/eora/eora.py
+++ b/gptqmodel/eora/eora.py
@@ -13,7 +13,7 @@
 #   year={2024}
 # }
 
-from typing import Sequence, Tuple
+from collections.abc import Sequence
 
 import torch
 from torch import Tensor
@@ -29,7 +29,7 @@ def eora_process_input(
         name: str,
         sample_size: int,
         device: torch.device,
-) -> Tuple[int, torch.Tensor, float]:
+) -> tuple[int, torch.Tensor, float]:
     """Prepare the per-batch covariance contribution required for EoRA.
 
     The contribution remains on the originating device so multi-GPU execution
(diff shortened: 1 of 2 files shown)


Qubitium:
I will merge this first to gpt-qmodel. If issues arise, we can fix in another pr.

PR #3050, opened 2026-09-03
support lm-head/embed requant

ZX-ModelCloud:


ZX-ModelCloud:
PR Review
PR: ModelCloud/GPTQModel#3050
Reviewed head SHA: 319aee222a5bff92c06bfeb7c3ec9b136e490247
Result: findings

Findings:
1. Severity: medium
   File and line: optimize/requant_embed_lm_head.py:85
   Rationale: On a tied-weight model, get_output_embeddings() is the same object as the input embedding, and named_modules() normally resolves both endpoints to the input path. The CLI therefore deduplicates the output target here and installs the requested bits/group-size override only for the input path. ModuleLooper unties the modules later, producing a new lm_head path; that path then receives the hard-coded 8-bit/group-32 fallback instead of the requested output configuration. Output-only and both modes can silently quantize the output endpoint with the wrong settings.
   Action: Untie tied endpoints before resolving target_names, or explicitly carry the requested output override from the pre-untie alias to the new output module name. Add a tied-model test that asserts the effective config on both runtime paths.
2. Severity: high
   File and line: gptqmodel/models/writer.py:291
   Rationale: Single-endpoint requantization cannot save a tied source checkpoint correctly. In output-only mode, matched_shards_by_prefix contains only the output prefix, so the fallback condition requiring the input prefix in that mapping is false and save raises at line 304 because tied checkpoints commonly omit lm_head.weight. In input-only mode, only the input replacement is written, but line 785 persists tie_word_embeddings=false after ModuleLooper cloned an untied output head; the cloned lm_head weight is never serialized, so reload has a missing output weight.
   Action: Resolve the original tied input shard independently of the selected replacement prefixes and serialize the untied output state whenever the saved config becomes untied, including input-only mode. Cover input-only and output-only saves from a source containing only the tied input tensor, then reload both outputs.
3. Severity: medium
   File and line: gptqmodel/nn_modules/qlinear/__init__.py:1549
   Rationale: The embedding branch leaves int_weight with one row per vocabulary entry, but the packing loop always reads a full pack_factor rows from every ceil-sized qweight row. For vocabularies not divisible by pack_factor, such as GPT-2 with 50,257 tokens at 4 bits, the final iteration indexes beyond int_weight and crashes. The current unit test uses a five-row embedding but stops before packing, while the end-to-end checkpoint happens to have an aligned vocabulary.
   Action: Pad the embedding rows and corresponding g_idx to the packing width or bound and zero-fill the final partial word, and ensure dequantization preserves or slices to the original vocabulary size. Add a pack/save/reload/forward test with a non-divisible vocabulary.

Qubitium:
Reviewed head `319aee222a5bff92c06bfeb7c3ec9b136e490247`.

**Result: changes recommended — three correctness findings.** This review confirms two issues from the earlier discussion and adds the calibrated `embed_only` issue.

1. **[P1] Single-endpoint requantization does not preserve a tied source checkpoint.**  
   [writer.py:291–304](https://github.com/ModelCloud/GPTQModel/blob/319aee222a5bff92c06bfeb7c3ec9b136e490247/gptqmodel/models/writer.py#L291-L304)  
   When the source has `tie_word_embeddings=true` and stores only the input embedding weight, output-only saving cannot use the fallback: `matched_shards_by_prefix` contains only selected replacements, so the required input prefix is absent. It raises `ValueError: Could not find checkpoint tensor for embedding module `lm_head`.` Input-only saving has the complementary problem: the looper creates an independent output head and the writer saves `tie_word_embeddings=false`, but only the input replacement is serialized; the source's omitted output weight remains missing. The resulting checkpoint cannot faithfully reload the runtime model.

   Resolve the tied source shard independently of the replacement selection, and include the retained output weight when untying an input-only checkpoint. Add input-only and output-only save/reload coverage using a source that omits the tied output tensor. The new lifecycle test covers only BOTH.

2. **[P2] Embedding packing reads past vocabularies that do not fill a packing word.**  
   [qlinear/__init__.py:1535–1549](https://github.com/ModelCloud/GPTQModel/blob/319aee222a5bff92c06bfeb7c3ec9b136e490247/gptqmodel/nn_modules/qlinear/__init__.py#L1535-L1549)  
   The new embedding branch keeps one `int_weight` row per token, while the existing packing loop unconditionally reads `pack_factor` rows for every ceil-sized output word. At 4 bits/int32, a vocabulary of 17 reaches row 17 and raises `IndexError`; 50,257 rows fails the same way. The embedding class advertises arbitrary input dimensions and automatic padding, but this path does not pad the source rows.

   Zero-fill or safely bound partial words, and keep unpacking/`g_idx` consistent with the original vocabulary length. Add a pack/save/reload/lookup test with an unaligned vocabulary, rather than stopping at the floating-point quantization result.

3. **[P2] The calibrated embedding path silently ignores `embed_only=False`.**  
   [stage_layer.py:545–548](https://github.com/ModelCloud/GPTQModel/blob/319aee222a5bff92c06bfeb7c3ec9b136e490247/gptqmodel/looper/stage_layer.py#L545-L548)  
   `quantize()` now forwards `QuantizeEmbedConfig` to `ModuleLooper`, but the looper retains only `embed_quant_mode`. This condition consequently disables decoder subset planning whenever any embedding mode is present, even for a fresh unquantized model requested with `QuantizeEmbedConfig(..., embed_only=False)`. Decoder blocks are replayed without quantization, although the operation ultimately marks the model quantized. The weight-only looper already distinguishes this flag.

   Carry `embed_only` through the calibrated flow and suppress decoder quantization only for the appropriate embedding-only operation. Test a fresh small model with `embed_only=False`, asserting that both the requested endpoint and decoder projections are quantized.

**Validation:** inspected the 19-file diff and surrounding quantization, packing, and serialization code. Executed the actual writer helper functions extracted from this commit with in-memory tensor/I/O doubles: OUTPUT raised the missing-prefix error, INPUT wrote only the input replacement, and BOTH wrote both replacements. Executed the actual NumPy allocation/packing loop: vocabulary 16 passed; 17 and 50,257 raised the out-of-bounds error. These are focused logic reproductions, not full checkpoint integration tests. Full PyTorch/CUDA tests were unavailable because PyTorch is not installed in this environment.

**Clarification on the earlier CLI finding:** tying weights does not by itself make the input and output modules the same object. Distinct embedding/linear modules can share a weight parameter while retaining different runtime paths. The earlier path-deduplication claim needs an actual module-aliasing model to establish its applicability; I am not counting it as a general tied-weight defect here.

ZX-ModelCloud:
Follow-up remediation — all three findings from the validation review are addressed in [022994c3](https://github.com/ModelCloud/GPTQModel/commit/022994c394a37495fa1a89bac1d2bba9973d5ae3).

**Tied embedding replacement save lifecycle.** Fixed single-endpoint requantization for tied source checkpoints. Input-only saves now serialize the cloned untied output head and preserve the untied config; output-only saves locate and rewrite the tied input shard correctly. Added regression coverage for both, input-only, and output-only endpoint selection.

**Unaligned embedding vocabulary packing.** Fixed 2/4/8-bit embedding packing when the vocabulary size is not divisible by the packing width (for example, vocab size 17 or GPT-2 vocab size 50257). Packing now pads the tail safely, while dequantization trims the packed tail back to the original vocabulary size. Added forward and safetensors save/reload coverage.

**Embedding quantization stage selection.** Preserved decoder subset plans when `embed_only=False`, so GPTQ/calibrated embedding-plus-decoder quantization still runs the decoder quantization stages. Embedding-only mode continues to use decoder replay only. Added regression coverage for the setting propagation and subset-stage execution.

Regression suite: `100 passed, 13 skipped` (CUDA-only skips). Targeted Ruff checks and `git diff --check` also pass.

Qubitium:
Merge conflict

ZX-ModelCloud:
> Merge conflict

Conflict resolved.

PR #3052, opened 2026-09-04
Shared-input Hessian dedup: plan metadata, E2E validation, and telemetry

Qubitium:
## Summary

Model-definition metadata that identifies decoder-layer modules consuming the *same* input activation, and looper wiring so the GPTQ Hessian (`H = XᵀX`) is collected once per group and copied to the other members (e.g. `q/k/v_proj`, `gate/up_proj`).

Phase 1 (8bf42b7, e7cfd32): metadata + CPU probe. Phase 2 (e7ec666): looper dedup. Review round 2 (latest commit): per-model_type verification gate + same-subset `dedup_count`.

Contract (after review): **dedup is explicit opt-in only, and only for real-forward-verified model types.** Every quantizable module is a singleton group unless sibling leaves under the same parent carry the same `:in=<tag>` *and* the definition lists `model_config.model_type` in its own (non-inherited) `shared_input_verified_model_types`. Subset digits are never used to infer sharing. Runtime dedup only happens between group members in the same subset block.

## What Changed

- New `gptqmodel/models/shared_input.py`:
  - `:in=<tag>` module_tree leaf flag (ordinary flag to existing parsers; does not alter emitted blocks).
  - `build_shared_input_plan(module_tree, layer_modules) -> SharedInputPlan` with frozen `SharedInputGroup(key, parent, modules, subset_indices, explicit)`; helpers `leader_for`, `followers_of`, `shares_input`, `is_explicit`, `for_subset`, `filter_modules`, `with_prefix`, `dedup_count`.
  - Grouping: untagged -> singleton (`key == module path`); same parent + same `:in=<tag>` -> shared (`key == "<parent>:in=<tag>"`). `:!`/`:?` leaves excluded; `experts.{i}` expand per expert (routed experts never share with each other). `build_shared_input_plan(..., explicit_tags=False)` ignores all tags (singletons only).
  - Same-subset runtime semantics: `SharedInputGroup.dedup_followers` = members that share a subset block with an earlier member (one leader per block); `dedups_at_runtime`, `members_in_subset(i)`; `SharedInputPlan.dedup_count = sum(len(g.dedup_followers))`. A group spanning blocks (qwen3.5 `in_proj_qkv:0`/`in_proj_z:1`, glm4_moe nested `shared_experts`) is still probe-verified but contributes 0 — matching what `GPTQProcessor` actually does. `followers`/`leader` remain the structural view.
  - `collect_leaf_specs()` raises `ValueError` when a leaf is repeated across `module_tree` variants with different `subset_tag` / `input_tag` / `quantize`; identical duplicates are fine.
  - `probe_shared_inputs(layer, plan, forward) -> SharedInputProbeReport`: pre-hooks every planned module, runs a real forward, compares captured inputs (shape/dtype/values/call-count). Status: `has_errors` (mismatches or undeclared identical inputs), `fully_verified` (no errors, no missing modules, no uncalled groups), `ok == fully_verified` (strict).
- `BaseQModel.shared_input_plan(model_config, quantize_config, is_awq_quantize=False)` classmethod, gated by `BaseQModel.shared_input_verified(model_config)`:
  ```python
  shared_input_verified_model_types: frozenset[str] = frozenset()   # BaseQModel default, NOT inherited
  def shared_input_verified(cls, model_config):
      return model_config.model_type in cls.__dict__.get("shared_input_verified_model_types", ())
  def shared_input_plan(cls, model_config, ...):
      return build_shared_input_plan(cls.module_tree, layer_modules, explicit_tags=cls.shared_input_verified(model_config))
  ```
  `module_tree` is inherited (every Llama clone), the verification set is not, so DeciLM/Dream/Ernie4.5/InternLM/Instella/MobileLLM/Xverse/... and any `model_type` merely mapped onto `LlamaQModel` stay singleton-only until they get a real-forward case. `model_config=None` -> unverified.
- Verified sets: `LlamaQModel {llama, mistral, gemma, granite, olmo2, stablelm, cohere, cohere2}`, `Qwen2QModel {qwen2}`, `Qwen3QModel {qwen3}`, `Gemma2QModel {gemma2}`, `Gemma3QModel {gemma3_text}`, `Phi3QModel {phi3}`, `Qwen3MoeQModel {qwen3_moe}`, `Qwen2MoeQModel {qwen2_moe}`, `MixtralQModel {mixtral}`, `DeepSeekV3QModel {deepseek_v3}`, `GLM4MoEGPTQ {glm4_moe}`, `Qwen3_5TextQModel {qwen3_5_text}`, `Qwen3_5_MoeTextQModel {qwen3_5_moe_text}`, `Qwen3NextGPTQ {qwen3_next}`, `GPTOSSGPTQ {gpt_oss}`, `Llama4TextQModel {llama4_text}` — exactly the `CASES` of the CPU forward suite. The image-text wrappers (`Qwen3_5QModel`, `Qwen3_5_MoeQModel`, `Llama4QModel`, `Gemma3ForConditionalGeneration`) keep their tags but are **not** verified (singleton plans) until covered.
- MLA definitions (deepseek_v2/v3/v32/v4/vl_v2, glm4_moe_lite, glm5_next, glm_moe_dsa, kimi_k25, longcat_flash, minicpm3, axk2) keep `q_b_proj:…:in=q_a`, `kv_b_proj:…:in=kv_a` as documented singletons — the probe showed they receive different tensors (`(2,6,16)` vs `(2,1,6,16)`).
- README section under "How to Add Support for a New Model".
- `full_layer_modules()`/`simple_layer_modules()` output unchanged for all definitions (`:in=` is transparent to block emission).

### Phase 2 — looper Hessian dedup

- `HessianConfig.dedup_shared_inputs: bool = True` (serialized to quant meta; dynamic-overridable per module, e.g. `{".*\\.v_proj$": {"hessian": {"dedup_shared_inputs": False}}}`).
- `GPTQ.adopt_hessian_from(leader)`: `leader.materialize_global_hessian()` then a **private** fp32 copy of `H` (`copy=True`, follower's target device) plus `nsamples`/`fwd_counter`; clears follower partials. Raises on column mismatch. Independence matters because `quantize()` mutates `H` in place and `free()` drops it.
- `LoopProcessor.begin_shared_input_capture(model, subset_names, is_lm_head_module) -> {follower: leader}` / `end_shared_input_capture(subset_names)`: no-op defaults; `GPTQProcessor` overrides them:
  - plan derived once per model class via `model.shared_input_plan(model_config, quantize_config)`;
  - per subset pass, for each `explicit` group: candidates = members present in this subset, with a task of type exactly `GPTQ` (GPTAQ/FOEM excluded), per-module `hessian.dedup_shared_inputs` on, same `columns`; first in subset order leads, others follow; <2 candidates -> no dedup;
  - `pre_process_fwd_hook(name)` returns a no-op hook for followers (leader/singleton hooks unchanged, incl. keep-mask splitting);
  - `end_shared_input_capture` -> `follower.adopt_hessian_from(leader)`, marks follower as having captured input ids, clears election state; `shared_input_dedup_count` tracks adoptions.
- `stage_subset._run_single_subset_pass`: `begin_…` before hooks are installed (only when `execute_forward`), `end_…` right after hooks are removed and before coverage validation / worker quantization.
- Not touched: AWQ/QQQ/ParoQuant processors, MoE routing/early-stop, lm_head (never dedups), LoopProcessor lifecycle.
- Follow-up: opting more definitions in once they have real-forward coverage.

## Tests

`tests/module_tree/test_shared_input.py`: flag parsing, leaf-spec extraction incl. repeated nested keys, expert placeholder resolution, singleton default, explicit same-subset / cross-subset grouping, conflicting-variant metadata raises (tag, `:!`, `:?`, subset) while identical duplicates pass, `:!`/`:?` exclusion, block output unchanged with `:in=`, prefix/filter, concurrent determinism, MoE, every registered definition covers its quantizable paths and has no non-explicit shared group, synthetic probe pass/fail modes (shape, dtype, value, missing, uncalled, partial, multi-call, kwarg tensors, hook cleanup on success/exception, undeclared pairs, `has_errors`/`fully_verified`/strict `ok`).

`tests/module_tree/test_shared_input_cpu_forward.py`: tiny HF configs on CPU (hidden 32–64, 2 layers) through `convert_model` + real `model(input_ids)` probe for Llama, Qwen2/3, Mistral, Gemma/Gemma2/3, Granite, OLMo2, StableLM, Cohere/Cohere2, Phi3, Qwen2/3-MoE, Mixtral, DeepSeek-V3, GLM4-MoE, Qwen3.5 dense + MoE, Qwen3-Next, GPT-OSS, Llama4. Every layer must be `fully_verified` with `undeclared == ()` and every shared group `explicit`; negative tests show a wrong MLA tag -> `mismatches`, an untagged q/k/v triple -> `undeclared`, a wrongly split q/k -> `undeclared`, and un-routed experts -> `unverified` (strict `ok` False, `has_errors` False). `test_only_real_forward_verified_definitions_dedup[<every MODEL_MAP type>]` asserts: verified <=> has a real-forward case (same class), and unverified types yield `shared_groups == ()` / `dedup_count == 0`; `test_verified_model_types_are_not_inherited` checks a `LlamaQModel` subclass and an unlisted model_type stay singleton-only.

`tests/test_shared_input_hessian_dedup.py` (27, CPU): cross-subset group (`qkv:0`/`z:1`) drives the processor through all subset passes — no election, both members capture their own `H` (equal values, distinct storage), and `shared_input_dedup_count == plan.dedup_count`; `adopt_hessian_from` equal-but-independent `H` (mutating follower leaves leader intact), nsamples/fwd_counter carry-over, idempotent multi-follower, uncalled leader -> zero `H`, self no-op, column mismatch raises, `quantize()` after adopt == independent collection; election: first-in-subset leads, singletons/untagged never, subset-restricted members, missing tasks, `dedup_shared_inputs=False` (global + dynamic), lm_head, model without plan API, non-plain-GPTQ excluded, column mismatch skipped, state reset on next `begin`; capture: followers get 0 `add_batch` then adopt, hooks resume normal after `end`, dedup vs independent `H` allclose, keep-mask path, `end` without `begin` / with pruned tasks.

`tests/test_shared_input_hessian_dedup_e2e.py`: real `GPTQModel.load` + `quantize()` of a 2-layer tiny Llama (GQA, hidden 64) on CPU with dedup on vs off: elections are `{k,v -> q}` / `{up -> gate}` per layer, followers' `fwd_counter == 0`, and all 14 quantized weight tensors are bit-identical between the two runs.

```bash
cd tests && python -m pytest module_tree/test_shared_input.py test_shared_input_hessian_dedup.py module_tree/test_shared_input_cpu_forward.py -q
# 449 passed in 7.02s
cd tests && python -m pytest module_tree -q --deselect module_tree/test_moe_flag_parsing.py::TestMoEModuleName::test_get_moe_module_name_none_tree --deselect module_tree/test_subset.py::test_qwen3_5_moe_subset_early_stop_follows_module_tree_execution_order
# 333 passed, 14 skipped
cd format && ruff check --config ruff.toml ../gptqmodel/models/shared_input.py ../gptqmodel/models/base.py ../gptqmodel/models/definitions/*.py ../tests/module_tree/test_shared_input*.py
# All checks passed!
```

Pre-existing on `main` (unrelated, also fail without this branch): `test_moe_flag_parsing.py::TestMoEModuleName::test_get_moe_module_name_none_tree`, `test_subset.py::test_qwen3_5_moe_subset_early_stop…` (needs CUDA).

## Review Requirements

- [ ] I personally reviewed every file in this diff.
- [ ] I checked that the code matches existing project structure, APIs, and conventions.
- [ ] I avoided unnecessary monkeypatching and used the project's normal extension points where possible.

## Notes

Ready for review. Review round 1 (comments 5536991865 / 5536994939 / 5536996271) addressed in e7cfd32: singleton default + explicit opt-in, conflict detection across variants, strict `ok`. Review round 2 (5539747040): P1 -> per-model_type `shared_input_verified_model_types` gate + MODEL_MAP coverage test (+6 Llama-clone real-forward cases); P2 -> chose "scope to same-subset": `dedup_count`/`dedup_followers` now match the looper, README + tests updated. Cross-subset groups (qwen3.5 `in_proj_qkv:0`/`in_proj_z:1`) intentionally stay as-is: reusing the leader `H` across passes would only skip one `add_batch` while keeping `H` alive longer, so it is not worth the lifecycle complexity. CPU e2e smoke (load+quantize, dedup on/off bit-identical, save/reload/generate) in comment 5547829240.

Link to Devin session: https://app.devin.ai/sessions/62474a853aa048bf95738e1e03f06e20
Open in Devin Desktop: https://app.devin.ai/desktop/session/62474a853aa048bf95738e1e03f06e20?variant=devin
Requested by: @Qubitium


Qubitium:
### Review: P1 — do not infer shared-Hessian eligibility from subset grouping by default

The current default rule — **same immediate parent + same expanded quantization subset/block => same input tensor** — is not a safe semantic contract.

The numeric subset/block historically describes quantization/execution grouping, not tensor identity. This PR itself demonstrates the problem: MLA definitions had modules in the same subset that receive different activations, so they needed explicit `:in=` splits. That is evidence that subset membership is not sufficient proof of shared input.

The risk becomes correctness-critical in the follow-up that skips follower `add_batch`: a false-positive group will silently reuse the wrong `H = XᵀX` for another module and corrupt the quantization result without necessarily crashing.

The real-forward suite is good, but `CASES` currently validates only a relatively small subset of the registered model definitions. The all-definition test checks structural coverage/consistency; it does **not** prove that the inferred tensors are identical at runtime.

I recommend one of these safer contracts before wiring this into the looper:

1. **Safest:** default every module to a singleton and require explicit `:in=<tag>` opt-in for every dedup group.
2. Or add an explicit per-definition opt-in such as `infer_shared_inputs_from_subsets = True`, enabled only for definitions with real-forward verification.
3. At minimum, do not consume inferred/default groups for Hessian dedup until every dedup-enabled definition has runtime tensor-identity coverage.

The optimization is valuable, but the failure mode here is silent quantization error, so the default should be conservative rather than inferred from an unrelated grouping primitive.

Qubitium:
### Review: P1 — `collect_leaf_specs()` silently discards conflicting metadata across module-tree variants

`collect_leaf_specs()` currently does:

```python
if template in specs:
    return
```

so the **first definition wins** for a repeated relative leaf path across `module_tree` variants.

That is unsafe for `:in=` because `BaseQModel.build_layer_modules()` intentionally supports/merges multiple complete tree variants. If two variants contain the same relative module path but assign different `:in=` tags (or different quantize/subset metadata), the planner silently uses whichever variant was visited first.

The docstring even says the first definition wins "so variant trees cannot silently re-tag a shared leaf," but this implementation does the opposite: it silently hides the conflict.

Before this metadata can drive Hessian reuse, repeated leaf specs should be validated. If an existing spec differs in any semantic field (`input_tag`, `quantize`, and probably the relevant grouping metadata), raise a clear error rather than accepting the first value. Identical duplicate specs are fine.

Suggested shape:

```python
new = LeafSpec(...)
old = specs.get(template)
if old is not None and old != new:
    raise ValueError(f"conflicting module_tree metadata for {template}: {old} vs {new}")
specs.setdefault(template, new)
```

A unit test with two module-tree variants that repeat the same leaf with different `:in=` tags would catch this.

Qubitium:
### Review: P2 — `SharedInputProbeReport.ok` can report success when the plan was not actually verified

`ok` is currently:

```python
return not self.mismatches and not self.undeclared
```

so a probe with missing planned modules and/or entirely uncalled groups can still return `True`. The tests explicitly lock this behavior in (`test_missing_and_uncalled_modules`, `test_moe_unrouted_expert_is_reported_unverified`).

That makes `report.ok` easy to misuse as a validation gate. For a feature intended to prove Hessian-sharing safety, "no contradiction observed" is different from "verified." A missing module or uncalled group provides no evidence that its declared/inferred sharing is correct.

I would separate these states explicitly, for example:

- `has_errors`: mismatches / undeclared conflicts
- `fully_verified`: no errors **and** no missing modules **and** no unverified groups
- keep `ok` only if its meaning is documented as "no observed contradiction"

Or make `ok` strict and add a separate permissive property for MoE probes where unrouted experts are expected.

This matters because future CI or tooling is very likely to write only `assert report.ok`; with the current API that can silently bless an incompletely exercised plan.

Qubitium:
### Review summary

I reviewed the planner/probe implementation, the model-definition annotations, the targeted synthetic tests, the tiny real-forward coverage, existing PR discussion, and current CI state.

**Verdict: changes requested before this becomes a Hessian-dedup contract.** The implementation is well-tested structurally, but there are two correctness hazards that can turn into silent quantization error once follower Hessian collection is skipped:

1. Defaulting to `same parent + same subset => same input` relies on subset metadata that was not originally an input-identity guarantee, while only a subset of model definitions receive real-forward validation.
2. Conflicting `:in=` metadata across module-tree variants is silently resolved by first-definition-wins.

I also flagged the probe API semantics because `report.ok` can be true for missing/unverified groups, which is dangerous for future CI validation.

Current Ruff workflow is green; I did not find a CI failure driving these findings. The concerns are semantic/correctness issues in the new contract rather than formatting or test hygiene.

Qubitium:
@ZX-ModelCloud Validate

ZX-ModelCloud:
### Follow-up validation review

I reviewed the current head (`e7ec6669730e429401cf0a2c4bcc9623dd82ba3d`) and recommend **changes requested** before merge.

#### P1 — `LlamaQModel` tags implicitly enable dedup for unverified subclasses

The conservative contract introduced in this PR says that only model definitions verified through a real forward should carry `:in=` tags. However, the tags were added directly to `LlamaQModel.module_tree`, while `HessianConfig.dedup_shared_inputs` defaults to `True`.

Those tags are inherited automatically by definitions such as DeciLM, Dream, Ernie 4.5, InternLM, Instella, MobileLLM, Xverse, and other Llama-derived classes that do not override `module_tree`. Enumerating `MODEL_MAP` at this head shows **41 model-type entries with shared groups**, while the real-forward `CASES` suite contains only **17 cases**. Therefore, the statement that tags were added only to definitions exercised by the CPU suite is not currently true.

I additionally ran real-forward probes for Cohere, Cohere2, Gemma, Granite, OLMo2, and StableLM; all passed. That still leaves the trust-remote-code descendants and other inherited definitions unverified. Because an incorrect inherited tag silently skips follower Hessian capture and reuses the wrong `H`, I recommend either:

1. moving the opt-in metadata to each verified concrete definition, leaving unverified descendants disabled; or
2. adding real-forward coverage for every inherited model type that receives shared groups.

Relevant locations:

- `gptqmodel/models/definitions/llama.py:17-26`
- `gptqmodel/quantization/config.py:1214-1220`
- `tests/module_tree/test_shared_input_cpu_forward.py:64-186`

#### P2 — cross-subset groups are not deduplicated, and `dedup_count` overstates the result

The planner and README allow `:in=` groups to span subsets, including Qwen3.5's `linear_attn.in_proj_qkv:0` and `linear_attn.in_proj_z:1`. However, `GPTQProcessor.begin_shared_input_capture()` elects members only from the current `subset_names` and skips groups with fewer than two members in that pass. Consequently, a group with one member in each subset performs no Hessian dedup at runtime.

`SharedInputPlan.dedup_count` still counts every follower across the full group, so it reports a collection as skippable even when the looper will skip none. Either retain/reuse the leader Hessian across subset passes, or scope the API, count, metadata, and documentation to same-subset dedup.

Relevant locations:

- `gptqmodel/looper/gptq_processor.py:245-268`
- `gptqmodel/models/shared_input.py:227-229`
- `gptqmodel/models/definitions/qwen3_5.py:41-45`

#### Validation performed

- Exact PR head: `e7ec6669730e429401cf0a2c4bcc9623dd82ba3d`
- Targeted suite: **275 passed in 12.42s**
- `git diff --check`: clean
- All currently visible GitHub checks are green

The three earlier review issues are resolved correctly: untagged leaves now remain singletons, conflicting variant metadata raises, and `report.ok` is a strict alias of `fully_verified`. The same-subset Hessian adoption path is otherwise well structured: followers receive independent fp32 copies, so later in-place quantization and `free()` do not alias the leader's Hessian.

Qubitium:
**Re-review of head `08f797f52809464ae8b689de7035b7b7378b7541`: the P2 finding below is resolved. No new actionable findings in the follow-up diff.**

The new `_hessian_accumulation_settings()` guard compares each task's resolved `staging_dtype`, `chunk_size`, and `chunk_bytes` before electing it as a follower. Differing settings now leave that member on independent capture. Matching members can still share, so a mismatched k_proj does not prevent a compatible v_proj from following q_proj.

Validation of this update:

- Reviewed the complete delta from `48cc0d2d`: 14 implementation lines and 74 test lines, across two files.
- Executed the exact new settings helper and election method extracted from this head with lightweight fixtures. Independently varying **only** staging dtype, chunk size, or chunk byte budget excludes the mismatched member; matching non-default settings retain both followers. Per-module dedup disable also passes.
- Inspected the added low-precision regression, which compares dedup-on/off H and checks that the differently staged modules really have unequal Hessians.
- All currently returned GitHub check runs are successful.
- PyTorch remains unavailable in this environment, so the repository's PyTorch/real-forward/e2e suites and GPU execution were **not rerun** here.

The earlier review findings remain resolved. This closes my outstanding finding; the validation above does not establish GPU runtime parity.

<details>
<summary>Original review at 48cc0d2d — retained for context; P2 now resolved</summary>

Review of head `48cc0d2dc7398a77615f1d3b971f482647e19770`: **one remaining correctness finding (P2).**

### P2 — Respect per-module Hessian accumulation settings when electing followers

Location: [`gptq_processor.py:250–269`](https://github.com/ModelCloud/GPTQModel/blob/48cc0d2dc7398a77615f1d3b971f482647e19770/gptqmodel/looper/gptq_processor.py#L250-L269).

The election checks the explicit input group, exact GPTQ task type, `dedup_shared_inputs`, and input width, but does not check whether the tasks have compatible Hessian accumulation settings. Existing `clone_gptq_config_for_module()` supports per-module overrides of `hessian.staging_dtype`, `chunk_size`, and `chunk_bytes`.

For example, use a verified Llama q/k group with FP16 activations, `chunk_size=1`, BF16 staging on q_proj, and a dynamic override requesting FP32 staging on k_proj. Both tasks keep dedup enabled. The current election makes k_proj follow q_proj, so k_proj's capture hook is skipped and `adopt_hessian_from()` copies q_proj's BF16-staged statistics. The explicit FP32 override is silently ineffective.

This is a numerical difference, not just configuration bookkeeping: for one token with an FP16 input component `1.0009765625`, FP32 staging produces a normalized Hessian diagonal of `2.003908157348633`; BF16 staging first rounds that component to `1.0`, producing `2.0`. Copying the resulting H into FP32 storage cannot recover the lost precision. Thus dedup can change the follower's quantization relative to independent capture despite identical input tensors.

**Suggested fix:** partition each shared-input group by compatible effective Hessian accumulation settings, or conservatively leave members with differing settings on independent capture. Include chunking settings in that decision because they determine the accumulation path/order. Add a regression using low-precision activations and different per-module staging settings, comparing the follower's H with dedup disabled. The existing default-FP32 tests and dynamic dedup-disable test do not exercise this case.

### Previous findings

The current changes address the earlier reported issues: untagged leaves remain singletons; conflicting variant metadata raises; probe `ok` is strict; the model-type verification gate is non-inherited; and `dedup_count` now uses same-subset followers.

### Validation

- Reviewed the current diff, planner/probe, annotations, capture/adoption path, surrounding configuration/device lifecycle, new tests, and existing discussion.
- Executed the exact `begin_shared_input_capture()` method extracted from this head with lightweight task fixtures: it elects a follower despite different staging dtypes. Checked the numerical counterexample above using standard-library FP16/FP32 packing and BF16 rounding.
- All currently returned GitHub check runs are successful.
- PyTorch is unavailable in this review environment, so I did **not** rerun the repository's PyTorch/real-forward/e2e suites or GPU tests. The numerical example is an isolated arithmetic reproduction, not an end-to-end quantization run.

</details>

Qubitium:
## Validation update

This PR now includes full real-model E2E coverage and lifecycle telemetry for shared-input Hessian deduplication.

- Llama 3.2 1B Instruct: full CUDA quantize/save/reload/generate passed with `PYTHON_GIL=0`; 48 expected and 48 adopted deduplicated Hessians.
- Qwen1.5-MoE-A2.7B (smallest complete Qwen MoE found in `/monster/data/model`): full 24-layer / 4,488-projection quantize/save/reload/generate passed; 1,512 expected and adopted deduplicated Hessians.
- Free-threaded safety: 35 dedup tests passed with `PYTHON_GIL=0`, including concurrent plan derivation and concurrent follower adoption.
- Targeted regression sweep: 463 passed; focused Ruff checks and `git diff --check` passed.
- Lifecycle telemetry event: `hessian_input_collection_dedup`, with expected/adopted/cumulative counts, leader mapping, layer/subset identifiers, and `verified`/`mismatch` status. Enable structured records with `GPTQMODEL_DEVICE_TELEMETRY=1`.

Hardware note: the validation host exposes one `NVIDIA PG506-230` (96 GiB), not an A100 or dual-GPU host. The existing dual-GPU test therefore skips here; an actual A100/dual-GPU run remains required for hardware-specific certification.

PR #3071, opened 2026-09-09
refactor(machete): harden runtime cache

ZX-ModelCloud:
## Summary

- Move pinned CUTLASS checkouts and generated Machete sources into a versioned user cache.
- Verify the pinned CUTLASS archive with SHA256 before atomic publication; add safe extraction, file locking, completion manifests, and atomic cache publication.
- Support offline cache use, read-only configured CUTLASS checkouts, and an explicit precompiled-library path that never silently falls back to JIT.
- Keep generated artifacts out of the source distribution and document cache/prewarming controls.

## Validation

- Machete JIT tests: 34 passed, 3 CUDA skipped.
- Prebuilt-library tests: 3 passed.
- Real CUTLASS end-to-end generation: 10 sources.
- ruff, py_compile, and git diff --check: passed.
- Transparent note: the full tests/test_torch_ops_jit_extension.py suite still has one pre-existing Python ABI marker failure in logic not touched by this change; the new prebuilt-path tests pass.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -16,5 +16,6 @@ include gptqmodel_ext/floatx_cpu.cpp
 include gptqmodel_ext/marlin/generate_kernels.py
 include gptqmodel_ext/machete/generate.py
 recursive-exclude gptqmodel_ext __pycache__ *.pyc
+prune gptqmodel_ext/machete/generated
 prune tests/
 prune format/
--- a/README.md
+++ b/README.md
@@ -420,6 +420,21 @@ export GPTQMODEL_FP32_ACCUM=0
 
 JIT-compiled kernels are cached at `~/.cache/gptqmodel/torch_extensions` by default. Multiple processes on one host may safely share the cache: builds are serialized with a cross-process file lock that the OS releases automatically if a process dies.
 
+Machete's pinned CUTLASS checkout and generated CUDA sources are kept in the
+versioned user cache (`~/.cache/gptqmodel`, or `GPTQMODEL_CACHE_DIR`; when set,
+`XDG_CACHE_HOME/gptqmodel` is used). Set `GPTQMODEL_CUTLASS_DIR` to use an
+already-installed, read-only CUTLASS 4.7.1 checkout. `GPTQMODEL_OFFLINE=1`
+disables downloads and requires that checkout or a verified cache hit already
+exist. For environments where compilation is not allowed, point
+`GPTQMODEL_MACHETE_PRECOMPILED_LIBRARY` at a compatible Machete shared library;
+an invalid or missing explicit library is reported as an error and does not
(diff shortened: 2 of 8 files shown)


ZX-ModelCloud:
PR Review
PR: ModelCloud/GPTQModel#3071
Reviewed head SHA: a4a8ed1a3ab7c76b2326df99346ce31fb827d551
Result: findings

Findings:
1. Severity: medium
   File and line: gptqmodel/utils/cpp.py:1308
   Rationale: Compatibility is inferred only from the process-global torch.ops namespace after the configured library is loaded. If this namespace was already populated by an earlier JIT or prebuilt load, any unrelated but loadable shared library passes `_refresh_runtime_cache()`, and the old operators continue to be used even though the new path is recorded as authoritative. This is reproducible after changing the environment path in-process, including after `clear_cache()`, because PyTorch operator registrations are not unloaded.
   Action: Do not attribute pre-existing registrations to a newly configured artifact. Reject in-process path switches once the namespace is populated, or validate the artifact in an isolated process or with artifact-specific identity metadata before marking it loaded.
2. Severity: medium
   File and line: gptqmodel/utils/machete.py:470
   Rationale: Persisted-cache validation checks only the manifest-listed top-level `.cu` files and ignores every other directory entry. The generated sources use quoted includes for `machete_mm_launcher.cuh` and `machete_prepack_launcher.cuh`, so an extra header placed beside the cached `.cu` files shadows the trusted source-tree header during compilation while this integrity check still returns true.
   Action: Validate that the cache directory contains exactly the regular, non-symlink manifest-listed `.cu` files plus the marker and manifest, and/or emit angle-bracket launcher includes so the generated-source directory cannot shadow the trusted include root.

ZX-ModelCloud:
Addressed the previous review findings in commit `41021a9a1e4b17bdc0a34cab756f518d6c6fc6f1`:

1. Prebuilt torch.ops identity
   - The loader now rejects a configured prebuilt library when the required operators are already registered but cannot be attributed to that artifact.
   - It preserves the successfully loaded prebuilt path across `clear_cache()`, because PyTorch operator registrations cannot be unloaded in-process.
   - Switching to a different prebuilt path in the same process now fails with guidance to start a new process.

2. Generated Machete cache integrity
   - Cache validation now requires the generated directory to contain exactly the manifest-listed `.cu` files, the completion marker, and the manifest.
   - Every entry must be a regular non-symlink file, preventing an injected launcher header from shadowing the trusted source-tree header.

Regression tests cover pre-existing operator registrations, in-process prebuilt path changes, and shadow-header injection. Ruff checks and `git diff --check` passed; the relevant test files completed with 86 passed and 3 CUDA-only tests skipped. One unrelated ABI-marker test failure was verified as already present at the reviewed head.

Qubitium (approved):

PR #3072, opened 2026-09-10
fix: make kernel selection device-aware

ZX-ModelCloud:
## Summary

- preserve concrete torch.device ordinals through device-map normalization and kernel validation
- validate every target GPU and reject mixed accelerator families or compute capabilities explicitly
- make validation caching sensitive to device index, capability, dtype, shape, quantization, and training dimensions
- honor allow_marlin=False and move Marlin, Machete, Swordfish, and QQQ capability checks to the selected physical device
- isolate extension API tests from host-specific Hadamard and Swordfish availability

## Validation

- targeted Ruff checks passed
- CPU/mock regression: 143 passed, 62 skipped, 0 failed
- GPU regression on NVIDIA PG506-232 / SM80: 181 passed, 17 skipped, 0 failed
- Marlin FP16 and BF16 clean JIT builds passed
- two-GPU probe preserved cuda:0 and cuda:1 and queried both capabilities as 8.0

The remaining GPU skips are expected for optional backends and architectures unavailable on SM80, including Machete SM90 and Swordfish Blackwell runtime paths.

### DIFF
--- a/gptqmodel/extension.py
+++ b/gptqmodel/extension.py
@@ -79,7 +79,7 @@ _EXTENSION_SPECS = (
         name="machete",
         aliases=("gptq_machete", "awq_machete"),
         resolve=lambda: _resolve_extension_attr("gptqmodel.utils.machete", "_MACHETE_TORCH_OPS_EXTENSION"),
-        supported=lambda: _resolve_attr("gptqmodel.utils.machete", "_validate_machete_device_support")(),
+        supported=lambda: _resolve_attr("gptqmodel.utils.machete", "_validate_machete_build_support")(),
         unsupported_error=lambda: _resolve_attr("gptqmodel.utils.machete", "machete_runtime_error")(),
     ),
     _ExtensionSpec(
@@ -96,7 +96,7 @@ _EXTENSION_SPECS = (
         name="swordfish",
         aliases=("gptq_swordfish", "awq_swordfish"),
         resolve=lambda: _resolve_extension_attr("gptqmodel.utils.swordfish", "_SWORDFISH_TORCH_OPS_EXTENSION"),
-        supported=lambda: _resolve_attr("gptqmodel.utils.swordfish", "_validate_swordfish_device_support")(),
+        supported=lambda: _resolve_attr("gptqmodel.utils.swordfish", "_validate_swordfish_build_support")(),
         unsupported_error=lambda: _resolve_attr("gptqmodel.utils.swordfish", "swordfish_runtime_error")(),
     ),
     _ExtensionSpec(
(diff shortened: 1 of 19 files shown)


ZX-ModelCloud:
PR Review
PR: ModelCloud/GPTQModel#3072
Reviewed head SHA: efb5c08bcb9e7dd4852d3b5ab15c9f4a70581eaf
Result: findings

Findings:
1. Severity: medium
   File and line: gptqmodel/utils/importer.py:254
   Rationale: An unindexed accelerator value such as device="cuda" is normalized to torch.device("cuda"), then returned unchanged here because it is not a DEVICE enum. Kernel validation consequently probes only the current CUDA device, but from_quantized later builds its default layer map across torch.cuda.device_count(). On heterogeneous GPUs this can select Machete, Marlin, Swordfish, or QQQ based on cuda:0 and then place layers on an unsupported GPU, defeating the all-target validation introduced by this PR.
   Action: Expand unindexed torch.device accelerator values across all visible ordinals, or preserve them as DEVICE families until expansion, and add a regression test for device="cuda" with differing mocked capabilities.
2. Severity: medium
   File and line: gptqmodel/utils/importer.py:71
   Rationale: The same normalizer handles both Accelerate device-map entries and the public from_quantized device argument, so hard-coding every integer as a CUDA ordinal changes existing device=0 behavior. Previously normalize_device selected the available accelerator family, including XPU or NPU and CPU fallback; now an XPU-only or NPU-only host receives cuda:0 and follows the wrong kernel and placement path.
   Action: Interpret integers as CUDA ordinals only for device-map values, while resolving a public integer device against the active accelerator as before; cover device=0 on mocked XPU/NPU-only hosts.

ZX-ModelCloud:
Addressed the two review findings in commit `2bb9812a07d32006d14e724e77a985bb19913d49`.

- `gptqmodel/utils/importer.py:254`: unindexed accelerator devices such as `cuda` now expand to all visible ordinals before model-wide kernel validation and default layer mapping. Validation therefore checks every target GPU and rejects mixed compute capabilities instead of probing only the current device.
- `gptqmodel/utils/importer.py:71`: public integer device arguments retain the historical active-accelerator resolution for CUDA, XPU, NPU, and CPU fallback. Integer values inside Accelerate device maps remain CUDA ordinals; direct `select_quant_linear` calls preserve the same public behavior.
- Added regression coverage for unindexed CUDA expansion and public integer selection on mocked XPU/NPU hosts.

Validation: `pytest -q tests/kernels/test_selection.py` — 50 passed, 31 skipped; targeted NPU normalization tests — 1 passed, 1 skipped. Pinned Ruff project checks and `git diff --check` passed.

PR #3077, opened 2026-09-11
Validate required qweight tensors in GPTQ safetensors checkpoints

peter-positron:
## Summary

A GPTQ/GPTQ_V2 checkpoint missing a selected quantized layer's `qweight` can load and run with zero-initialized packed weights. Validate required storage before loading single-file and entirely safetensors-sharded checkpoints.

## What Changed

- Read actual shard headers so a stale index cannot conceal missing qweight in these inventories.
- Preserve tied dense heads, shared quantized module aliases, derived g_idx omissions and other stored formats; retain existing tying-helper defaults.

## Tests

- [x] I ran the new targeted test locally.
- [x] I ran other directly relevant local tests.

```bash
PYTHONPATH=. python -m pytest -p no:cacheprovider -q \
  tests/test_required_qweight.py tests/test_loader.py \
  tests/test_embedding_save_lifecycle.py
```

43 passed. Real local Llama GPTQ/GPTQ_V2 load-and-forward controls pass; six missing-storage cases expose the unpatched defect. Scoped Ruff and whitespace checks pass.

## Review Requirements

- [x] I avoided unnecessary monkeypatching and used normal extension points where possible.

## Notes

Only qweight presence is checked. Opaque or mixed inventories containing `.bin` shards retain existing behavior without this check; payloads are not deserialized for validation. This is not blanket strict state-dict loading. GPU backend conversions and the full suite were not exercised.

### DIFF
--- a/gptqmodel/models/loader.py
+++ b/gptqmodel/models/loader.py
@@ -82,6 +82,7 @@ from ..utils.model import (
     make_quant,
     no_placement_module_names,
     simple_dispatch_model,
+    validate_checkpoint_qweights,
 )
 from ._const import DEVICE, HAS_NPU, normalize_device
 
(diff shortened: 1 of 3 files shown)


Qubitium (approved):
LGTM. Code passing locally.

PR #3079, opened 2026-09-11
Apply model replay hooks during Paro layer preparation

peter-positron:
## Summary

Paro's generic replay preparation can add rotary embeddings to a Muse-Glimmer layer that uses no positional encoding. Apply the existing model hook before normalizing and caching prepared layer kwargs.

## What Changed

- Use the normal decoder replay extension point while preserving metadata normalization and layer-specific caching.
- Add real Muse-Glimmer output/gradient parity tests for grouped and streamed replay, with RoPE/NoPE and cache controls.

## Tests

- [x] I ran the new targeted test locally.
- [x] I ran other directly relevant local tests.

```bash
PYTHONPATH=. python -m pytest -p no:cacheprovider -q tests/test_paroquant.py \
  -k 'muse_glimmer_replay_matches_dense_forward_and_gradients or replay_hook_metadata_is_normalized_and_layer_specific'
PYTHONPATH=. python -m pytest -p no:cacheprovider -q tests/test_paroquant.py
```

New regressions: 8 passed; all 8 expose the unpatched behavior. Broader file: 118 passed, 8 skipped, 4 existing failures. The same four failures occur on the unchanged base (110 passed, 8 skipped) because existing processor fixtures lack `_input_cache_lock` before replay runs.

## Review Requirements

- [x] I avoided unnecessary monkeypatching and used normal extension points where possible.

## Notes

Tests pass scoped Ruff; source Ruff reports the same three baseline findings (import order, missing Sequence annotation import, unused variable). Existing source style is preserved. AWQ owner-layer routing and unrelated model signatures remain outside this decoder replay change. GPU and full-suite validation were not run.

### DIFF
--- a/gptqmodel/looper/paroquant_processor.py
+++ b/gptqmodel/looper/paroquant_processor.py
@@ -1079,6 +1079,14 @@ class ParoQuantProcessor(LoopProcessor):
                 module_kwargs["position_ids"] = _LayerShardLoader._tensor_to_device(position_ids, target_device)
 
         module_kwargs["use_cache"] = False
+        prepare_replay_kwargs = getattr(self.gptq_model, "prepare_layer_replay_kwargs", None)
+        if prepare_replay_kwargs is not None:
+            module_kwargs = prepare_replay_kwargs(
+                layer=layer,
+                layer_input=[x],
+                additional_inputs=module_kwargs,
+                target_device=target_device,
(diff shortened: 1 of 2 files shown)


Qubitium (approved):
Local tests passing. Bug validated and fixed. Thanks!

PR #3084, opened 2026-09-15
auto-decode native FP8/NVFP4 sources for W4A16

Qubitium:
## Summary
- detect NVIDIA ModelOpt FP8/NVFP4 source checkpoints and automatically inject BF16 AutoModuleDecoderConfig for GPTQ/AWQ
- validate BF16 and optional native FP8/NVFP4 linear support at load time; preserve packed source weights in checkpoint-backed storage
- use native FP8/NVFP4 wrappers only for the pre-quantization Hessian-collection forward; defer BF16 source decoding until the GPTQ/AWQ hand-off
- guarantee FP32 Hessian staging and accumulation for automatic floatx-source quantization; native source execution never downcasts Hessian math
- when native execution is unavailable or a wrapper cannot be built, continue with decoded BF16 and emit a warning plus structured telemetry
- emit floatx forward-plan, native/fallback-use, and deferred-decode telemetry events
- add a logged Torch-only NVFP4 decoder fallback when TorchAO is unavailable; require TorchAO 0.18.0

## Validation
- focused decoder/floatx suite: 27 passed, 14 deselected
- real Qwen3-8B-FP8 on GB10: full 36-layer native-FP8 -> GPTQ W4A16, group size 128 conversion completed; all 108 Hessian capture groups used the native FP8 path, BF16 decode was deferred, and Hessian accumulation remained FP32.
- saved W4 artifact reloads with TorchLinear on CUDA and scores against a native-FP8 teacher over 4 held-out prompts / 60 tokens: Top-256 support KL mean 0.17397, p95 0.96919, top-1 agreement 76.67%, teacher top-1 in student top-256 100%.
- calibration: 16 prompts concatenated to 256 tokens. This is a smoke-quality signal, not a broad benchmark.
- real NVIDIA Qwen3-8B-NVFP4 single-module route: native TorchFP4Linear Hessian forward, deferred BF16 source, FP32 Hessian partial, finite W4 GPTQ output (4096 x 12288).


Qubitium:
Reviewed head `9e7ecd81a4a222dc92748e65779d1467fd9a1d45`. **One high-priority integration failure and two medium-priority issues remain.** I would address these before merging.

### 1. [P1] Full NVFP4 conversion reaches dense layer materialization before the decoder

The new automatic source setup enables checkpoint-backed loading ([loader.py:244–257](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/models/loader.py#L244-L257)), but the normal quantization loop still calls `pre_quantize(module)` on the entire decoder layer before building/executing its subsets ([stage_layer.py:525](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/looper/stage_layer.py#L525)). For a meta shell, that calls `shell_module_materialize` with the default role. The packed-weight decoder interception only runs for `role="forward"` with a `named_module`; the default route recursively copies checkpoint tensors into dense shell parameters.

A ModelOpt NVFP4 weight has shape `[out, in/2]`, while the shell Linear expects `[out, in]`. LazyTurtle's ordinary layout transform cannot unpack it, so conversion fails before reaching the new per-module forward/deferred-decode path.

**Reproduced on CPU:** a meta layer containing `Linear(16, 8)` and a real safetensors checkpoint containing an `(8, 8)` uint8 weight plus scales fails through `BaseQModel.pre_quantize` with:

```text
LazyTurtle: submodule materialization param `0.weight` under `layer`:
checkpoint tensor could not be reshaped into the target layout
(checkpoint=layer.0.weight, source_shape=(8, 8), target_shape=(8, 16))
```

**Fix:** make layer-entry materialization decoder-aware: materialize the required ordinary tensors while leaving packed floatx linears for their per-module preparation, or decode them explicitly through the appropriate path. Add a one-layer test that enters through the regular quantization lifecycle. The reported NVFP4 single-module test bypasses this failure point.

### 2. [P2] Automatically injected BF16 decoder can conflict with FP16 model activations

The loader resolves and applies the model dtype before configuring floatx sources ([loader.py:907–934](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/models/loader.py#L907-L934)), then unconditionally injects `AutoModuleDecoderConfig(target_dtype=torch.bfloat16)` ([line 246](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/models/loader.py#L246)). It does not reconcile that choice with an explicit `dtype=torch.float16` or an FP16 source config retained by `auto_dtype`.

The decoded fallback is an ordinary Linear with BF16 parameters, without an input cast. FP16 hidden states therefore fail during fallback calibration; the same conflict can occur after a native FP8 module is replaced by its BF16 quantization source for replay.

**Reproduced on CPU:** the actual source-builder and forward-builder methods, given an FP16 shell and FP8 checkpoint weights, produce a BF16 Linear that raises:

```text
expected m1 and m2 to have the same dtype, but got: c10::Half != c10::BFloat16
```

The matching BF16 input succeeds.

**Fix:** reconcile the shell/model activation dtype with the decoder before model construction, or preserve a compatible execution dtype with explicit conversions at the boundary. Add coverage for FP16 source configs and explicit FP16 loading, including the decoded fallback and post-quantization replay.

### 3. [P2] “Native forward used” and deferred-decode telemetry do not establish native execution

The new `floatx_native_forward_used` event is emitted when a wrapper is selected, before its forward runs ([base.py:2715–2719](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/models/base.py#L2715-L2719)). In particular, the NVFP4 branch labels a constructed `TorchFP4Linear` as `validated_native_nvfp4` ([base.py:2783–2793](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/models/base.py#L2783-L2793)).

However, `TorchFP4Linear._native_weight()` constructs `NVFP4Tensor` without `act_quant_kwargs`. In the newly required **TorchAO 0.18.0**, that defaults to `None`, and [`nvfp4_linear` explicitly dequantizes the weight and calls ordinary `F.linear`](https://github.com/pytorch/ao/blob/v0.18.0/torchao/prototype/mx_formats/nvfp4_tensor.py#L581-L617). Thus this route expands BF16 weights during the Hessian forward; it is not evidence of a native FP4 GEMM or of all BF16 decoding being deferred until optimization.

FP8 has a related reporting gap: `TorchFP8Linear` only selects `_scaled_mm` for tensorwise scales and suitable shapes, and silently switches to dequantization on an execution error ([fp8.py:410–450](https://github.com/ModelCloud/GPTQModel/blob/9e7ecd81a4a222dc92748e65779d1467fd9a1d45/gptqmodel/nn_modules/qlinear/fp8.py#L410-L450)). Row/block wrappers can therefore also receive the native-use event while executing dense matmul.

**Fix:** distinguish wrapper/storage selection from the kernel actually executed, and emit execution/fallback events from the forward path. Keep the deferred *quantization-source allocation* claim separate from temporary forward dequantization. If actual FP4 GEMM is intended, configure and validate that dispatch explicitly, including its activation-quantization semantics.

### Validation

- Rechecked that the PR head remained `9e7ecd81a4a222dc92748e65779d1467fd9a1d45` before posting.
- Used an isolated CPU harness with the real LazyTurtle/safetensors materializer and source-extracted BaseQModel methods; package imports/checkpoint discovery were stubbed to avoid unrelated dependencies. Both failures above reproduced on PyTorch `2.14.0+cpu`.
- Checked all **256 packed-byte values** in the new Torch-only FP4 fallback against an independent E2M1 decode; values, nibble order, and signed zero matched.
- Traced the FP4 dispatch against the exact TorchAO `v0.18.0` source.
- Did **not** run the complete repository suite, GPU kernels, or an end-to-end model conversion. The PR's reported GPU results were read, not independently rerun.

PR #3092, opened 2026-09-18
Add caller-owned Marlin scratch buffers and bounded eager reuse

ZX-ModelCloud:
## Summary

This PR adds caller-owned temporary buffers for Marlin GEMM and an opt-in `MarlinScratchContext` for bounded reuse across sequential layers. The legacy allocation path remains unchanged when scratch is omitted, and GEMM outputs remain caller-independent fresh tensors.

## What changed

- Added optional caller-owned `c_tmp` and `a_tmp` buffers to the FP16/BF16 Marlin GEMM ABI and torch operators.
- Shared checked C++ capacity calculations between allocation and the sizing API.
- Added validation for scratch device, dtype, contiguity, capacity, alignment and forbidden aliases, using executed padded dimensions.
- Added `MarlinScratchContext` with bounded retained capacity, growth, partial fallback, explicit clear, device/stream/thread ownership and capture/reentrancy checks.
- Added CPU lifecycle/interface/capacity/regression tests, CUDA numerical tests and CUDA Graph replay tests.
- Fixed the AWQ scratch fixture to register its quantized buffers.
- Enabled and dispatched generated FP16/BF16 `kU8` kernels for AWQ 8-bit zero-point weights.

## Measured optimization result

Benchmark: `scripts/benchmark_marlin_scratch.py --graph`, 198 cases, `python_314t`, Torch `2.14.0+cu130`, one RTX 4090 (`cuda:0`) on a host with 4×RTX 4090. Final source commit: `36aa3a5f`.

- 191 cases completed; 7 cases reported strict bitwise-repeat differences in the existing narrow AWQ FP16 benchmark check (all are `narrow`, `K=4096`, `N=1024`; maximum absolute differences were below `5e-4`).
- Eager allocator requests per call: median `2 → 1` (50% fewer).
- `aten::empty` calls per audited trial: median `25 → 20` (20% fewer).
- Allocator-requested bytes per call: median `4.5 MiB → 256 KiB` (94.4% lower).
- Paired eager forward ratio (temporary/context): median `0.831`; the context path trades some dispatch overhead for the allocation reduction.
- CUDA Graph capture time: median `3.49 ms → 1.73 ms` with explicit scratch; replay latency remained effectively unchanged (`38.49 µs` vs `38.18 µs`).

The optimization therefore primarily reduces allocator pressure and graph-capture cost; it is not an unconditional eager-latency speedup.

## Follow-up review and comments

- Add concise comments explaining buffer ownership, capacity accounting, exception cleanup, mutable operator aliases and graph buffer lifetimes.
- Fix the benchmark's stale `MarlinScratchContext` import after the public-module change. A new CPU integration test reproduced the import failure before the fix and exercises the paired benchmark orchestration after it.
- Local follow-up acceptance: **90 passed, 87 skipped** in 13.86 seconds, using the documented test command including `tests/test_marlin_scratch_benchmark.py`. Ruff and `git diff --check` passed.
- The follow-up environment has no available CUDA device, so the new local benchmark run records 198 `not_run` cases. The separate GPU measurements above are preserved; no new GPU results are claimed for this follow-up revision.


ZX-ModelCloud:
PR Review
PR: ModelCloud/GPTQModel#3092
Reviewed head SHA: 36aa3a5f401d6e0067c7e4bc72feab733f35f121
Result: no-findings

Findings: none

